PR #1, opened 2021-03-14
embed icons

jneilliii:
quick fix for embedding the icons rather than linking to github images


borisbu (approved):

borisbu:
Thanks, my idea was to show that the user can customize  the icons with images from everywhere without havingto upload them to the raspberry pi, but smart users will figure that out i guess ;-)

jneilliii:
It will still allow them to use whatever they want, just the initial load of the images after install will use those internal addresses and not have to go out to the internet to load unless they want it to.

PR #11, opened 2021-03-27
Add OS commands

sydrbloc:
Added a field to execute os.system commands. 
Purpose is to optionally run a system command with a GPIO pin or instead of a GPIO pin
Feature would allow to use curl to trigger a remote gpio through a remote web api


katanabladesman:
> 

Could you add to the wiki for how to set up?

PR #16, opened 2021-04-05
Settings retouch

RobinTail:
1. Moved label on top.
2. Display label as a hint for disabled relays.
3. "ON on boot" moved to the group "Auto ON/OFF".
4. Delay setting is visible only when the "OFF after printing" is enabled.
5. Icons preview.
6. Minor naming changes.

image


borisbu:
Thanks a lot.. You are a big help

PR #32, opened 2021-08-04
added monitored state

skobkars:
In case when multiple scripts/plugins work with GPIO pins
monitored state allows to skip action if it is already in
desired state.

Only UI update is ran in that case.

Also, update_ui checks if model has been defined already to prevent errors in logs


skobkars:
New one is coming with GPIO states polling. Please disregards this one

PR #33, opened 2021-08-05
added GPIO polling thread

skobkars:
Added GPIO polling thread to correctly show
pin statuses when multiple scripts/plugins
can change them.


skobkars:
Replaced ResettableTimer with RepeatedTimer. Simplifies things a bit

Issue #25, opened 2021-05-01
`OctoRelayPlugin` object has no attribute `model`


XAMCTBO:
Hi
after latest update is stopped working (octopi  1.6.0)
octopi (updated)
PI updated

log shows
`2021-05-01 15:33:10,480 - octoprint.plugins.pluginmanager - INFO - Loaded plugin repository data from disk, was still valid
2021-05-01 15:33:10,495 - octoprint.plugin - ERROR - Error while calling plugin octorelay
Traceback (most recent call last):
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint/plugin/__init__.py", line 271, in call_plugin
    result = getattr(plugin, method)(*args, **kwargs)
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint/util/__init__.py", line 1890, in wrapper
    return f(*args, **kwargs)
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint_octorelay/__init__.py", line 220, in on_event
    self.update_ui()
  File "/home/pi/oprint/lib/python3.7/site-packages/octoprint_octorelay/__init__.py", line 312, in update_ui
    self.model[index]['iconText'] = iconOn
AttributeError: 'OctoRelayPlugin' object has no attribute 'model'
2021-05-01 15:33:10,687 - octoprint.server.util.flask - INFO - Passively logging in user alex from ::ffff:192.168.1.28`

RobinTail:
The supplied log shows that `on_event` called before `on_after_startup` which sets the `model`.
This issue has most likely been already resolved on September 2021 in this commit:

https://github.com/borisbu/OctoRelay/commit/d9e9f93a37c2cb87104d50047246e88f94f41145#diff-515c8cdc2c54449d864cbd805fc587c8d46e167902e3b462c16925371d0a3c2cR264

Please retry with most recent versions again.
If the issue remains, please create another issue with a similar log.
Thank you, @XAMCTBO

PR #44, opened 2021-12-05
`@OCTORELAY` command support

### LINKED ISSUE #43: Feature: @command support
It came up today on the Discord server that having @ command support in the plugin may be beneficial for users of OctoDash. I'm going to look into submitting a PR for this feature, but will require a little refactor I think since all the work is currently happening in the SimpleApiPlugin mixin implementation. Will break out the actual working bits of the function to another function that can be called by the API, which would then also allow for integrating the @ command support.

jneilliii:
adds `@OCTORELAY r#` command support for use in OctoDash to toggle state changes, closes #43


RobinTail:
@jneilliii , please resolve conflicts for the PR

octoprint_octorelay/__init__.py
```diff
@@ -200,83 +206,90 @@ def on_shutdown(self):
...
-        # added api command to get led status
-        if command == "getStatus":
-            return flask.jsonify(status=ledState)
-            
-        else:
             self._logger.debug("Ocotrelay before pin: {}, inverted: {}, currentState: {}".format(
                 relay_pin,
                 inverted,
-                ledState
+                led_state
```
RobinTail, line 230: Consider detaching the feature introduction from the refactoring.
That would make your PR much more compact (less changes).
Also, `ledState` is a legacy from a plugin which OctoRelay was based on (see Readme).
The naming might be is subject for changes in another PR.

octoprint_octorelay/__init__.py
```diff
@@ -15,6 +15,7 @@
 
 POLLING_INTERVAL = 0.3
 
+
```
RobinTail, line 18:
```suggestion
```

less changes

octoprint_octorelay/__init__.py
```diff
@@ -74,7 +80,7 @@ def get_settings_defaults(self):
                 autoOffDelay=10,
             ),
             r4=dict(
-                active = True,
+                active=True,
```
RobinTail, line 83:
```suggestion
                active = True,
```
less changes

octoprint_octorelay/__init__.py
```diff
@@ -289,10 +302,10 @@ def on_event(self, event, payload):
...
         elif event == Events.PRINT_FAILED:
             self.print_stopped()
-        #elif event == Events.PRINT_CANCELLING:
-            # self.print_stopped()
-        #elif event == Events.PRINT_CANCELLED:
-            # self.print_stopped()
+        # elif event == Events.PRINT_CANCELLING:
+        # self.print_stopped()
+        # elif event == Events.PRINT_CANCELLED:
+        # self.print_stopped()
```
RobinTail, line 308:
```suggestion
        #elif event == Events.PRINT_CANCELLING:
            # self.print_stopped()
        #elif event == Events.PRINT_CANCELLED:
            # self.print_stopped()
```
less changes

octoprint_octorelay/__init__.py
```diff
@@ -414,10 +430,11 @@ def input_polling(self):
                 self.update_ui()
                 break
 
+
```
RobinTail, line 433:
```suggestion
```
less changes

octoprint_octorelay/__init__.py
```diff
@@ -354,30 +365,35 @@ def update_ui(self):
...
-                self.model[index]['confirmOff'] = confirmOff
+                self.model[index]['iconText'] = icon_on
+                self.model[index]['confirmOff'] = confirm_off
             else:
-                self.model[index]['iconText'] = iconOff
+                self.model[index]['iconText'] = icon_off
                 self.model[index]['confirmOff'] = False
 
-        #self._logger.info("update ui with model {}".format(self.model))
+        # self._logger.info("update ui with model {}".format(self.model))
```
RobinTail, line 389:
```suggestion
        #self._logger.info("update ui with model {}".format(self.model))
```
less changes

RobinTail (changes requested):
OK. Now the diff is a bit more readable.

- [x] Tests require adjustments.
- [ ] New functions should be tested
- [ ] Missing `pin` as a command param is most likely a bug that should be fixed separately
- [ ] Replacing `==` with `is (not) True/False` is a refactoring due to [best practices](https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true), preventing type casting — I presume the whole module requires global review on applying that practice. Perhaps it's also worth it to extract into another.
- [x] I removed changes of variable renames — we should install a linter for this and introduce a rule on naming convention. Another refactoring to do later.
  - [x] made #95

octoprint_octorelay/__init__.py
```diff
@@ -229,27 +233,33 @@ def on_api_command(self, command, data):
...
-        settings.update(self._settings.get([index]))
+        if command == "update":
+            status = self.update_relay(data["pin"])
+            return flask.jsonify(status=status)
 
-        relay_pin = int(settings["relay_pin"])
-        inverted = settings['inverted_output']
-        cmdON = settings['cmdON']
-        cmdOFF = settings['cmdOFF']
+    def update_relay(self, index, get_status=False):
```
RobinTail, line 245: I'm leaning against the approach:
the function called `update_relay` being called with `get_status` parameter does NOT update the relay (lines 260-261)

octoprint_octorelay/__init__.py
RobinTail: Made an issue for that #94

octoprint_octorelay/__init__.py
```diff
@@ -202,21 +207,20 @@ def on_shutdown(self):
 
     def get_api_commands(self):
         return {
-            "update": [],
-            "getStatus": [],
-            "listAllStatus":[],
+            "update": ["pin"],
+            "getStatus": ["pin"],
+            "listAllStatus": [],
```
RobinTail, line 212: yeah, this is a bug, according to the documentation:
https://docs.octoprint.org/en/master/plugins/mixins.html#simpleapiplugin
parameters must be listed here.
But that should be fixed in another PR

RobinTail: Created #96 for this

RobinTail: PR #97

RobinTail: fixed

octoprint_octorelay/__init__.py
RobinTail: fixed in 1ab39eac9dabad030ef136d16df0124629763259

octoprint_octorelay/__init__.py
```diff
@@ -26,6 +26,11 @@ class OctoRelayPlugin(
     octoprint.plugin.RestartNeedingPlugin
 ):
 
+    def __init__(self):
+        self.polling_timer = None
+        self.model = None
+        self.turn_off_timers = None
```
RobinTail, line 32: this might fix #25

octoprint_octorelay/__init__.py
```diff
@@ -28,6 +28,11 @@ class OctoRelayPlugin(
     octoprint.plugin.RestartNeedingPlugin
 ):
 
+    def __init__(self):
+        self.polling_timer = None
+        self.model = None
```
RobinTail, line 33: I'm not sure about safety of this statement because of this issue in the past:
https://github.com/borisbu/OctoRelay/issues/25

The presence of `model` property (which is set by `on_after_startup`) is a condition involved in `on_event`:

```py
    def on_event(self, event, payload):
        self._logger.debug("Got event: {}".format(event))
        if event == Events.CLIENT_OPENED:
            if hasattr(self, 'model'):
                self.update_ui()
```

RobinTail: It seems that in some cases, `on_event` can be triggered before `on_after_startup`

RobinTail: fixed in 68ecafd6cd0a697aa9bf99ae6ac9966ea9214210 by constructing as dictionaries

octoprint_octorelay/__init__.py
```diff
@@ -226,15 +228,14 @@ def has_switch_permission(self):
...
+        self._logger.debug("on_api_command {}, some_parameter is {}".format(command, data))
 
         # added api command to get led status
         if command == "listAllStatus":
             GPIO.setwarnings(False)
             activeRelays = []
             for key in self.get_settings_defaults():
-                settings = self.get_settings_defaults()[key]
-                settings.update(self._settings.get([key]))
+                settings = self._settings.get([key], merged=True)
```
RobinTail, line 238: now it's time for me to figure out what `merged=True` does...

RobinTail: ok, it seems that this feature only affects the requests having more than 1 item in the setting path array

https://github.com/OctoPrint/OctoPrint/blob/f027aaa7300f70b046525a49245b33c43dd198c1/src/octoprint/settings/__init__.py#L1647-L1650

RobinTail: tested the consistency in 77343f4c09c8d72efc4857c06b2043fa9a5cfbf4

octoprint_octorelay/__init__.py
```diff
@@ -363,7 +370,7 @@ def print_stopped(self):
     def turn_off_pin(self, relay_pin, inverted, cmdOFF):
         GPIO.setup(relay_pin, GPIO.OUT)
         # XOR with inverted
-        GPIO.output(relay_pin, inverted != False)
+        GPIO.output(relay_pin, inverted is not False)
```
RobinTail, line 373: regarding this.
this syntax is really better since it's strict, not type casting.
However, neither the method `turn_off_pin` nor its caller do ensure the type of `inverted` parameter as boolean.
Meaning that we actually need a more loose comparison here until the type will be guaranteed.

RobinTail:
```suggestion
        GPIO.output(relay_pin, inverted != False)
```

octoprint_octorelay/__init__.py
```diff
@@ -373,7 +380,7 @@ def turn_off_pin(self, relay_pin, inverted, cmdOFF):
     def turn_on_pin(self, relay_pin, inverted, cmdON):
         GPIO.setup(relay_pin, GPIO.OUT)
         # XOR with inverted
-        GPIO.output(relay_pin, inverted == False)
+        GPIO.output(relay_pin, inverted is not True)
```
RobinTail, line 383:
```suggestion
        GPIO.output(relay_pin, inverted == False)
```

same reason here

octoprint_octorelay/__init__.py
```diff
@@ -247,29 +248,36 @@ def on_api_command(self, command, data):
...
-        else:
+
+        if command == "update":
             if not self.has_switch_permission():
                 return flask.abort(403)
+            status = self.update_relay(data["pin"])
+            return flask.jsonify(status=status)
+
+    def update_relay(self, index):
+        try:
```
RobinTail, line 268: I'm trying to figure out that can be catched here...

octoprint_octorelay/__init__.py
```diff
@@ -408,6 +415,11 @@ def update_ui(self):
         #self._logger.info("update ui with model {}".format(self.model))
         self._plugin_manager.send_plugin_message(self._identifier, self.model)
 
+    def process_at_command(self, comm_instance, phase, command, parameters, tags=None, *args, **kwargs):
+        if command == "OCTORELAY":
+            index = parameters
+            self.update_relay(index)
+            return None
```
RobinTail, line 422: Documentation

> Handlers are expected to return one of the following result variants:
> None: Don’t change anything. 

https://docs.octoprint.org/en/master/plugins/hooks.html#sec-plugins-hook-comm-protocol-atcommand-phase

RobinTail (approved):
OK. I'm finally happy with it

RobinTail:
QA passed ✅

Issue #46, opened 2021-12-20
Resets when print starts


GordonMKF:
Setup relay to switch printer off and on. Works fine until print loaded and starts then printer is shutoff.
Setup a relay on a second Pi with another printer but only controlling a light and same situation. On print start light switches off. Print continues and I can switch the light on without any issues.

![image](https://user-images.githubusercontent.com/17933476/146827643-97903181-2718-4dca-a851-317a60874035.png)

Printer on/off relay is externally powered so only signal triggering. Light relay on second Pi is powered from Pi.
Raspi 3B+'s with all updates done.

Ruggero1912:
Same here

mArJaNu-yV:
Same situation. orctoprint 1.7.3.

NDR008:
I got the same issue... :/

borisbu:
For my relay board, I need to invert the output. Did you try to do that?

![image](https://user-images.githubusercontent.com/10959304/153159095-d976187f-7b97-4798-9956-c2fee5e843db.png)

NDR008:
@borisbu I am using a different module that was design for the ESP D1 mini.
But basically your program is working for turning on / off when I click the icon.
It just turns off when I click print.

Question - is the GPIO being used in pull-down or pull-up? Are you setting the internal resistor?

NDR008:
@borisbu I understood what you mean... if you use the relay the opposite way, drive the relay to keep the printer off, the bug does not occur... but there must be a reason for this.
Let me try to poke at the code myself :)

NDR008:
@borisbu Update again: If I edit this line:
https://github.com/borisbu/OctoRelay/blob/5a940f5fa253930341db3ae0e864bb691fb39a5e/octoprint_octorelay/__init__.py#L320
From
GPIO.output(relay_pin, inverted != True)

to
GPIO.output(relay_pin, inverted == True)
The relay works as expected without Inverted settings
So somewhere here something is bizarre...

NDR008:
I made a pull request that should fix it.

Diaprojector:
Hi, I've had this issue as well (came here to report it). Seems like this issue is only on the GPIO23 pin. Changed it to GPIO17 and had no issues.

jshep321:
Hi,
[...]

PR #48, opened 2022-02-09
Fixes for non-inverted relay

NDR008:
I created a new turn_on_pin function
and called that function when print_started.

~~Also I made sure print_started is only called when that channel is active, because some people might have devices connected to relays that they de-activated.~~


NDR008:
@borisbu I think this fixes the bug

NDR008:
This change is not merged yet. Did you install my branch?

storbi81:
I installed everything via the plugin-manager from Octoprint.

NDR008:
> Good morning!
> 
> I think this change is responsible for a new problem that I have now with my printer. On print_started, both of my relais switched to on now, but I can't change this anywhere.
> 
> A few days ago, i tried to install the plugin Octoprint-Smart-Filament-Sensor from Anni L., but it will not work with OctoRelay (only the standard R1-R4 will be displayed in Navbar, nothing worked, you can configure what you want, no change). I've deinstalled both of the plugins and reinstalled OctoRelay again. I think, now I have some trouble with the different versions.

This change is not merged yet. Did you install my branch? How did you confirm that this change caused the issue?
What do you mean multiple-versions?

NDR008:
> I installed everything via the plugin-manager from Octoprint.

Then you did not use this change, because this change is not yet part of the publicly available plugin from the plugin manager.

NDR008:
> I created a new turn_on_pin function and called that function when print_started. Also I made sure print_started is only called when that channel is active, because some people might have devices connected to relays that they de-activated.

Thinking about your problem, I think this problem is actually one of the things I fixed in my change.
Are you able to load my code to test it? you have to manually download my python file and substitute it on your OctoPi/Octoprint.

ArmEagle:
Thank you. This fixed the exact problem you described in #46 that I was having myself too after adding a relay to control the power to the printer. And luckily it's very easy to add a modified plugin to OctoPi.

Zbysekz:
Confirmed - this solves problem when controlling printer and lights by GPIO outputs normally works, but if you start printing it shuts down the printer. Thanks!

RobinTail:
@ArmEagle && @Zbysekz && @NDR008 ,

do I understand it correctly, that this PR fixes the issue #46 ?

NDR008:
In my opinion yes.  I've been using it over a year.

octoprint_octorelay/__init__.py
```diff
@@ -349,6 +346,16 @@ def turn_off_pin(self, relay_pin, inverted, cmdOFF):
         self._logger.info("pin: {} turned off".format(relay_pin))
         self.update_ui()
 
+    def turn_on_pin(self, relay_pin, inverted, cmdON):
+        GPIO.setup(relay_pin, GPIO.OUT)
```
RobinTail, line 350: @NDR008 , is it ok, that exactly the same command is being executed right before calling this function, in line 318?

https://github.com/borisbu/OctoRelay/blob/cd52345b03b61de12cfd26eb275bf42d0441f47c/octoprint_octorelay/__init__.py#L318

octoprint_octorelay/__init__.py
```diff
@@ -349,6 +346,16 @@ def turn_off_pin(self, relay_pin, inverted, cmdOFF):
...
+        # XOR with inverted
+        GPIO.output(relay_pin, inverted == False)
+        GPIO.setwarnings(True)
+        if cmdON:
+            os.system(cmdON)
+        self._logger.info("pin: {} turned on".format(relay_pin))
+        self.update_ui()
```
RobinTail, line 357: @NDR008 , the function `turn_on_pin` is being called within a cycle introduced in line 309:

https://github.com/borisbu/OctoRelay/blob/cd52345b03b61de12cfd26eb275bf42d0441f47c/octoprint_octorelay/__init__.py#L309

Which makes this statement (updating the UI) being fired for each of the relays in configuration.

Apart from that, the same UI update is fired in line 320 anyway, after finishing the cycle:

https://github.com/borisbu/OctoRelay/blob/cd52345b03b61de12cfd26eb275bf42d0441f47c/octoprint_octorelay/__init__.py#L320

Question: is there any benefit of firing the UI update within the cycle? Can we keep the single UI update in the end of the cycle?

RobinTail (changes requested):
We've recently solved #55 that filters out deactivated relays as well.
Please adjust you implementation by resolving the conflicts, @NDR008

RobinTail:
Conflict resolved by preserving the solution on only active relays handling in target branch that was fixed in another PR: #55.

octoprint_octorelay/__init__.py
RobinTail:
```suggestion
```

removing UI update here, since it's called in the end of `print_stopped` which calls this method in a cycle

octoprint_octorelay/__init__.py
```diff
@@ -313,12 +313,12 @@ def print_started(self):
             relay_pin = int(settings["relay_pin"])
             inverted = settings['inverted_output']
             autoONforPrint = settings['autoONforPrint']
+            cmdON = settings['cmdON']
             active = settings["active"]
             if autoONforPrint and active:
                 self._logger.debug("turning on pin: {}, index: {}".format(relay_pin, index))
                 GPIO.setup(relay_pin, GPIO.OUT)
```
RobinTail, line 320:
```suggestion
```

removing this line here, since:
- same one exists in the `turn_on_pin` which is called in the next line
- for consistency with `turn_off_pin` method doing the same

octoprint_octorelay/__init__.py
RobinTail: made a suggestion to keep only this one

RobinTail (commented):
OK. Now I like its appearance.

RobinTail:
The commands I used to update this PR based on the latest changes:
```
gh pr checkout 48
git merge develop
git push git@github.com:NDR008/OctoRelay.git HEAD:master
```

RobinTail:
All tests succeeded, but `os.system` has to be mocked for `test_print_started`:

Screenshot 2023-07-23 at 12 29 41

RobinTail:
Sweet!

Screenshot 2023-07-23 at 13 37 27

octoprint_octorelay/__init__.py
```diff
@@ -333,8 +332,6 @@ def print_stopped(self):
             cmdOFF = settings['cmdOFF']
             active = settings["active"]
             if autoOFFforPrint and active:
-                self._logger.debug("turn off pin: {} in {} seconds. index: {}".format(
-                    relay_pin, autoOffDelay, index))
```
RobinTail, line 337: not sure why this was removed.
it does not seem to be related to the fix

octoprint_octorelay/test__init.py
```diff
@@ -307,6 +307,20 @@ def test_turn_off_pin(self, systemMock):
             systemMock.assert_called_with("CommandMock")
         restoreUI()
 
+    @patch('os.system')
+    def test_turn_off_on(self, systemMock):
```
RobinTail, line 311: wrong naming

```suggestion
    def test_turn_on_pin(self, systemMock):
```

RobinTail (approved):
👍🏽

RobinTail:
QA passed

Issue #52, opened 2022-05-25
ON before printing is "stuck" activated, despite NOT being activated in settings!


IAmOrion:
<img width="707" alt="Screenshot 2022-05-26 at 00 44 30" src="https://user-images.githubusercontent.com/3501380/170387285-918149d4-5dd0-4d12-9626-6d1715c62809.png">

As you can see in my screenshot, "ON before printing" is _NOT_ ticked/active.  Yet, every time I start a print, the relay turns on! 
With my relay board, the inverted output is required.  For all intents and purposes it works as expected if I manually toggle the relay control.  It also correctly turns OFF when a print is finished as expected.  I just don't want it turning on when a print starts... but for some reason it does.

shooter-z:
Can confirm. Octorelay 1.2.0 running on raspberry pi.
It is not a display bug, since behaviour is the same whenever "ON before printing" is enabled or disabled.

IAmOrion:
I'm assuming this plugin is dead/no longer supported. :(

hcooper:
I've not tested your exact problem, however I fixed an issue with "on before printing" not working for me.

https://github.com/borisbu/OctoRelay/pull/55

The issue was that the `print_started()` function, was processing all relay configs, not just the "enabled ones". At but chance the default configs were trying to use the same pin as me.

From your screenshot I see you're using pin 23. That's the same pin the default config for Relay 4 uses. Until my PR is committed, try changing Relay 4 config to a different/fake pin and see if that fixes your problem.

IAmOrion:
@hcooper - well done discovering that! That fixed the problem for me, thanks for the info.

RobinTail:
🚀  @IAmOrion , the issue should be fixed in v1.4.1.
Please check it out, and in case it's not, please create another issue.

Issue #39, opened 2021-10-26
Not switching on at start of print job


FlyingPete:
Hi, I have a small Prusa i3Mk3S+ print farm, all running OctoPrint with relay controlled nights on each printer.
I have tested with the OctoRelay plugin but I cannot get the switch on at start of print job to function, the switch off at completion works fine.
I have also tested with the Enclosure plugin, and switch on at start of print job works with that plugin.

mattdy:
I'm also experiencing this after updating to OctoPrint 1.7.2

Auto-off after print completion still appears to work, but auto-on is no longer functioning.

schneemaier:
I have the same issue. I usually print through Cura, but now i have connect to octoprint an manualy turn on the printer.

FalcoonM:
Issue is still present I guess. Turning off after a print job works flawlessly, but relay does not engage at start. Might this be related to how upload is handled. New print just show on a list and are unable to start if printer is not connected. I have just a data cable - no 5V between Pi and board, don't kno if this would affect the issue.

tonysprenk:
I'm having the same issue

hcooper:
I just filed https://github.com/borisbu/OctoRelay/pull/55 which fixes this for me.

The cause for me was that octorelay is processing other unactive configs too. In my case I was using pin 18, however there was an unactive config for "relay 3" which also used pin 18 which had `inverted_output=True`. So my pin 18 would go on, and then get turned off again straight away by the other config.

Until the fix is committed, if you use pins that conflict with any of the defaults listed below you might want to go in a put fake pin numbers in these configs:
```
R1 4
R2 17
R3 18
R4 23
R5 24
R6 25
R7 8
R8 7
```

ninthclowd:
> Until the fix is committed, if you use pins that conflict with any of the defaults listed below you might want to go in a put fake pin numbers in these configs:
> 
> ```
> R1 4
> R2 17
> R3 18
> R4 23
> R5 24
> R6 25
> R7 8
> R8 7
> ```

thanks for the workaround until #55 is merged!

RobinTail:
🚀  @FlyingPete , the issue should be fixed in v1.4.1.
Please check it out, and in case it's not, please create another issue.

Issue #18, opened 2021-04-08
Request - AutoConnect


Lonelyeagledk:
Hi

Is it possible to make an automatic connection to the printer from OctoPrint after OctoRelay have powered the printer on. When I power off the printer from OctoRelay the printer is total without power and the connection from OctoPrint are closed.
Thank you :)

Kind Regards
Mads

rich-dobbs-13440:
Hi folks!

I just installed the plugin today and have a similar request.  When I reboot Octopi with my current configuration, the printer is powered on, but the connection with the printer has failed.  It succeeds at that point if I click the button to connect.  But it seems like the interface should allow specifying a time to wait before Octopi attempts to connect with the printer.  Let me know if the current architecture of Octopi makes this difficult for you to implement.

Best regards,

Rich Dobbs

Ano10:
Hi,
Autoconnect is a good idea. Connect printer after relay is activated would be convenient.
In a same way autodisconnect would be great. Disconnect printer before relay is deactivated would avoid "SerialException" error from octoprint.

Best regards,
Arnaud

RobinTail:
That could be really a good feature. I like this idea.

RobinTail:
I found out that the following event is being fired after switching the printer's relay on:
```
2023-08-01 22:49:26,133 - octoprint.plugins.octorelay - DEBUG - on_api_command update, parameters {'pin': 'r2', 'command': 'update'}
2023-08-01 22:49:26,230 - octoprint.plugins.octorelay - DEBUG - OctoRelay before pin: 12, inverted: True, relay_state: False
2023-08-01 22:49:27,276 - octoprint.server - INFO - Serial port list was updated, refreshing the port list in the frontend
2023-08-01 22:49:27,289 - octoprint.plugins.octorelay - DEBUG - Got event: ConnectionsAutorefreshed
```

RobinTail:
Looks like connection can be implemented similar to EnclosurePlugin

https://github.com/vitormhenrique/OctoPrint-Enclosure/blob/6e35aa59409a71392a341af9a6a54f10cde1df53/octoprint_enclosure/__init__.py#L1789

```py
self._printer.connect()
```

It's not clear when this property or mixin is available.

RobinTail:
I checked it, the prop is actually available for the plugin:

```
octoprint.plugins.octorelay - INFO - start OctoRelay
[...]

PR #55, opened 2022-11-28
Make `print_started()` only process "active" relay configs.

hcooper:
This fixes an issue with relays not being turned on when a print starts.

I use OctoRelay to control only 1 relay (Pin #18). However there are 7 other placeholder/default configs. One of these placeholder configs happens to also be configured for Pin #18, and has `inverted_output: True` set. After `print_started()` turns on my relay, it then processes the placeholder config and flips it off again.

The debug logs show the conflicting pin configs:
```
settings for r1: {'active': True, 'relay_pin': '18', 'inverted_output': False,
settings for r2: {'active': False, 'relay_pin': 17, 'inverted_output': True,
settings for r3: {'active': False, 'relay_pin': 18, 'inverted_output': True,
[....]
```
_Side note: There seems to be some type confusion. Inactive configs have relay_pin as a int, but active configs have it as a string._

This PR basically makes `print_started()` follow the same logic of `print_stopped()`:
 - Filter for only `active` relays. **[the main fix]**
 - Add debug log statement when actually flipping a pin. **[helpful improvement]**
 - Move definition of`relay_pin` and `inverted` so both functions match. **[code tidy up]**

This should fix these two issues:

- https://github.com/borisbu/OctoRelay/issues/39
- https://github.com/borisbu/OctoRelay/issues/52

_Side note: "active" is not a great choice of name for this variable, as it's similar in meaning to "turned on", which is confusing when managing switches and configs. I think "enabled" would be better, as something can be enabled, but not active. But that's something to fix another time._

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -315,7 +315,7 @@ class OctoRelayPlugin(
             autoONforPrint = settings['autoONforPrint']
             active = settings["active"]
             if autoONforPrint and active:
-                self._logger.debug(f"turning on pin: {relay_pin}, index: {index}")
+                self._logger.debug("turning on pin: {}, index: {}".format(relay_pin, index))
                 GPIO.setup(relay_pin, GPIO.OUT)
                 # XOR with inverted
                 GPIO.output(relay_pin, inverted != True)
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -310,11 +310,12 @@ class OctoRelayPlugin(
             settings = self.get_settings_defaults()[index]
             settings.update(self._settings.get([index]))
 
+            relay_pin = int(settings["relay_pin"])
+            inverted = settings['inverted_output']
             autoONforPrint = settings['autoONforPrint']
-            if autoONforPrint:
-                relay_pin = int(settings["relay_pin"])
-                inverted = settings['inverted_output']
-
+            active = settings["active"]
+            if autoONforPrint and active:
+                self._logger.debug(f"turning on pin: {relay_pin}, index: {index}")
                 GPIO.setup(relay_pin, GPIO.OUT)
                 # XOR with inverted
                 GPIO.output(relay_pin, inverted != True)


octoprint_octorelay/__init__.py
```diff
@@ -310,11 +310,12 @@ def print_started(self):
             settings = self.get_settings_defaults()[index]
             settings.update(self._settings.get([index]))
 
+            relay_pin = int(settings["relay_pin"])
+            inverted = settings['inverted_output']
             autoONforPrint = settings['autoONforPrint']
-            if autoONforPrint:
-                relay_pin = int(settings["relay_pin"])
-                inverted = settings['inverted_output']
```
RobinTail, line 316: @hcooper , is there any reason for moving these two assignments from the scope of the condition up to the higher scope?

RobinTail: Ah, I presume this is the answer from the PR description.

> Move definition ofrelay_pin and inverted so both functions match. [code tidy up]
> makes print_started() follow the same logic of print_stopped()

So, it's for achieving similarity to the implementation of another method, right?

octoprint_octorelay/__init__.py
```diff
@@ -310,11 +310,12 @@ def print_started(self):
...
-
+            active = settings["active"]
+            if autoONforPrint and active:
+                self._logger.debug(f"turning on pin: {relay_pin}, index: {index}")
```
RobinTail, line 318: this syntax required Python 3.6 and it has not been used in the currently existing code.
[PEP 498](https://www.python.org/dev/peps/pep-0498/)

Is it OK for the plugin compatibility, @hcooper ?

RobinTail: For some reason, the plugin declares the support of Python 2.7.
But probably this is another issue that should be fixed.

Screenshot 2023-07-17 at 23 12 11

RobinTail: Well, the compatibility is declared here:

https://github.com/borisbu/OctoRelay/blob/b08a0e4a1d9f38114e312acfc09f704929d4c349/octoprint_octorelay/__init__.py#L417

RobinTail: Changing the Python compatibility must be a major version bump according to SemVer.
I'd like to keep this PR on a patch level.
In this regard, I'd like to request rewriting this line in a similar way it's implemented in `print_stopped` method (using `.format()` instead of `f""`),
@hcooper

RobinTail:
```suggestion
                self._logger.debug("turning on pin: {}, index: {}".format(relay_pin, index))
```

RobinTail (changes requested):
Please adjust logging implementation before we fix the Python compatibility in #74 , @hcooper

hcooper:
Accepting changes without testing (octopi defaulted to py3 three years ago).

RobinTail:
Good to know. Thank you very much.

RobinTail (approved):

RobinTail:
### 🚀  `1.4.1rc1` released

Regression test 🆗

PR #68, opened 2023-07-02
JS: Types constraints and tests

RobinTail:
Contributes to #65 (in part of Javascript).

- Developing frontend scripts in Typescript.
- Generating the final Javascript file.
- Testing frontend scripts.

```
 PASS  ./octorelay.spec.ts
  OctoRelayViewModel
    ✓ Should push the model into the registry (4 ms)
    ✓ Constructor should set its certain properties (1 ms)
    ✓ Message handler should ignore other recipients
    ✓ Message handler should process the supplied configuration (2 ms)

--------------|---------|----------|---------|---------|-------------------
File          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
--------------|---------|----------|---------|---------|-------------------
All files     |     100 |      100 |     100 |     100 |
 octorelay.ts |     100 |      100 |     100 |     100 |
--------------|---------|----------|---------|---------|-------------------
Test Suites: 1 passed, 1 total
Tests:       4 passed, 4 total
Snapshots:   6 passed, 6 total
Time:        0.299 s, estimated 1 s
```

⚠️ There is another piece of Javascript inside a template. I'd like to move it from there and test as well.
For that one I made another PR: #72

Issue #65, opened 2023-07-02
We need tests


RobinTail:
Dear community,

I love this plugin and I'd like to keep it alive. I'm going to maintain it further, but I need your help.

I am a frontend engineer and my primary focus is Typescript, Javascript, Node.js and React.

My knowledge of Python is very limited. But I also love 3D printing and use this plugin every day.

Luckily, I have a huge experience of software development and maintenance, based on which I realise that: 

- **We need tests** to continue adding new features and even fixing the bugs.
  - This step is crucial in order to avoid making new bugs to the existing implementation.
  - Ideally, each function should be covered by [unit tests](https://realpython.com/python-testing/).
- We also need a simple CI workflow that would run them for each commit.
  - The example of it you may find below.
  - It should address both `master` and `develop` branches.

Therefore, if you (yes you, %username%) are reading this being familiar with writing unit tests in Python, consider making a PR featuring the tests and a CI workflow for running them. Your contribution will be highly appreciated.

Suggestions, objections and concerns in this regard are welcome in comments to this issue.

### Example CI workflow

```yaml
# .github/workflows/ci.yml
# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python

name: Python package

on:
  push:
    branches: [ "master" ]
  pull_request:
    branches: [ "master" ]

jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.8", "3.9", "3.10"]

    steps:
    - uses: actions/checkout@v3
    - name: Set up Python ${{ matrix.python-version }}
      uses: actions/setup-python@v3
      with:
        python-version: ${{ matrix.python-version }}
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        python -m pip install flake8 pytest
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
    - name: Lint with flake8
      run: |
        # stop the build if there are Python syntax errors or undefined names
        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
    - name: Test with pytest
      run: |
        pytest
```

RobinTail:
I found an example of OctoPrint plugin tests:
[...]

PR #72, opened 2023-07-04
Moving JS from template

RobinTail:
It's better to keep JS out from the template.
It's easy in our case, since the only piece of it is actually a click handler.

This PR is an addition to #68 and features the complete tests coverage of all our UI code.

The PR contributes to #65


octoprint_octorelay/templates/octorelay_navbar.jinja2
```diff
@@ -1,28 +1,7 @@
...
-        $("#octorelay-confirmation-dialog").modal("show");
-    }
-</script>
-
 {% for n in range(1,9) %}
     <div id="relaisr{{n}}" style="float: left; padding: 8px; font-size: 1.3em; {% if n > 4 %} display:none; {% endif %}">
         <a
             id="ralayIconr{{n}}"
-            onclick='OctoRelayConfirmClick({"pin":"r{{n}}"}, this.title, this.getAttribute("data-confirm"))'
-            data-confirm="false"
```
RobinTail, line 25: there is no longer need for this

RobinTail:
🚀 1.4.0rc3, QA passed

PR #80, opened 2023-07-20
JS: changing the target level to ES6.

RobinTail:
Increasing the JS language level to ES6 (ES2015).

FYI, OctoPrint 1.9.x is using ES9 (ES2018).

⚠️ Potentially breaking change requiring users to upgrade their browser

Minimal versions of the different browsers supporting ES6 are listed here:
https://caniuse.com/?search=es6


RobinTail:
✅ operational

PR #82, opened 2023-07-20
2.0: Fix `img` nodes shaking

RobinTail:
Restoring icons dimensions on their consuming side.
Restoring SVG initial dimensions in `pt`.
This prevents IMG nodes shaking when operating the controls.
The issue revealed during the QA of 2.0.0rc1.


RobinTail:
Fixed ✅

PR #89, opened 2023-07-23
Improving and refactoring the tests

RobinTail:
- Removing excessive statements from the tests
- Ensuring that some methods are not called conditionally
- Testing the exception handling
- Coverage `100%`
- Clarification for the CI tests naming
- Testing against Python up to `3.11`
- Testing against OctoPrint down to `1.8` (compatibility)

PR #90, opened 2023-07-23
CI: ensure testing against OctoPrint 1.8

RobinTail:
It turned out that #89 did not actually force to install OctoPrint 1.8.
This PR should fix this.

Tested versions:
- OctoPrint 1.8: on Python 3.7–3.9
- OctoPrint 1.9: on Python 3.7-3.11


RobinTail (commented):
Yeah!

PR #97, opened 2023-07-24
Listing the required API param `pin`.

### LINKED ISSUE #96: Bug: API commands listed without required params
See `get_api_commands` method. First two should list the required `pin` parameter.

this is a bug, according to the documentation:
https://docs.octoprint.org/en/master/plugins/mixins.html#simpleapiplugin
parameters must be listed here.

_Originally posted by @RobinTail in https://github.com/borisbu/OctoRelay/pull/44#discussion_r1272748418_

RobinTail:
The issue originally fixed by @jneilliii in #44 
I'm cherry-picking it into another PR in order to release the fix separately.
Closes #96

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -202,9 +202,9 @@ class OctoRelayPlugin(
 
     def get_api_commands(self):
         return {
-            "update": [],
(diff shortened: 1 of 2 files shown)


RobinTail (commented):
OK

PR #99, opened 2023-07-25
Prohibit operating relays for readonly users

### LINKED ISSUE #51: Control Buttons Not Being Hidden From Read Only Users
When logging in as a read only user the OctoRelay icons still show in the status bar and I am able to toggle the light.  I have not tried toggling the printer power as I am in the middle of a print, but I would assume it would power off the printer also.  It appears the plugin does not adhear to the access groups set up in OctoPrint.  Please advise / fix.  I am on OctoRelay v1.2.0 and OctoPrint v1.7.3  Let me know if you need any other info.

RobinTail:
Closes #51 

Approach: introducing a custom permission.

- Initially enabled for Admins and Users (Operators).

Constraints:

- Avoid showing the buttons
- Reject the API command `update`
- Avoid toggling relays having "ON initially on boot"
- Avoid toggling relays having "Auto ON/OFF before/after printing"

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -336,7 +336,7 @@ class OctoRelayPlugin(
             autoONforPrint = settings['autoONforPrint']
             cmdON = settings['cmdON']
             active = settings["active"]
-            if autoONforPrint and active:
+            if autoONforPrint and active and self.has_switch_permission():
                 self._logger.debug("turning on pin: {}, index: {}".format(relay_pin, index))
                 self.turn_on_pin(relay_pin, inverted, cmdON)
         self.update_ui()
@@ -352,7 +352,7 @@ class OctoRelayPlugin(
             autoOffDelay = int(settings['autoOffDelay'])
             cmdOFF = settings['cmdOFF']
             active = settings["active"]
-            if autoOFFforPrint and active:
+            if autoOFFforPrint and active and self.has_switch_permission():
                 self._logger.debug("turn off pin: {} in {} seconds. index: {}".format(
                     relay_pin, autoOffDelay, index))
                 self.turn_off_timers[index] = ResettableTimer(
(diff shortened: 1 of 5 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -424,10 +425,22 @@ def input_polling(self):
                 self.update_ui()
                 break
 
+    def get_additional_permissions(self):
+        return [{
+            "key": "SWITCHING",
+            "name": "Switching relays ON and OFF",
+            "description": "Allows to toggle the GPIO pins and execute OS commands associated with them",
+            "roles": [ "default" ],
```
RobinTail, line 433: I'm not sure what this one should be.
Documentation here: https://docs.octoprint.org/en/master/plugins/hooks.html#octoprint-access-permissions

RobinTail: Here is how to use it:

```
    def on_api_get(self, request):
        if not Permissions.PLUGIN_{CODE}_{KEY}.can():
            return flask.abort(403)
```

https://github.com/OctoPrint/OctoPrint-PiSupport/blob/7fcdeaec0cb663523942dd017e25a79a3daaab69/octoprint_pi_support/__init__.py#L398C1-L400C36

octoprint_octorelay/__init__.py
```diff
@@ -250,6 +258,9 @@ def on_api_command(self, command, data):
             return flask.jsonify(status=ledState)
             
         else:
+            if not self.has_switch_permission():
+                return flask.abort(403)
```
RobinTail, line 262: this statement has to be tested

RobinTail (commented):
OK. Now I'm happy with it.

RobinTail:
It turned out that all events are not happening in the user's context.
So all `on_` functions can not utilise the permission.

PR #100, opened 2023-07-26
Relay switching permission

### LINKED ISSUE #51: Control Buttons Not Being Hidden From Read Only Users

RobinTail:
Closes #51 
Continuation of #99 

It turned out that `on_` events are happening outside the user context, so they don't have access to Permission.

### DIFF
--- a/octoprint_octorelay/test__init.py
+++ b/octoprint_octorelay/test__init.py
@@ -288,6 +288,7 @@ class TestOctoRelayPlugin(unittest.TestCase):
                     "confirmOff": False
                 }
             self.plugin_instance.update_ui()
+            permissionsMock.PLUGIN_OCTORELAY_SWITCH.can.assert_called_with()
             for index in self.plugin_instance.get_settings_defaults():
                 self.plugin_instance._settings.get.assert_any_call([index])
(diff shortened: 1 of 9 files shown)


octoprint_octorelay/test__init.py
```diff
@@ -253,7 +261,7 @@ def test_update_ui(self):
         GPIO_mock.input = Mock(return_value=False)
         cases = [
             { "inverted": True, "expectedIcon": "ON" },
-            { "inverted": False, "expectedIcon": "OFF" }
+            { "inverted": False, "expectedIcon": "OFF" },
```
RobinTail, line 264:
```suggestion
            { "inverted": False, "expectedIcon": "OFF" }
```

RobinTail (commented):
OK. Let's try this one

RobinTail:
QA passed ✅

PR #112, opened 2023-07-27
Ref: extracting constants

### LINKED ISSUE #103: Default settings should be extracted into static const
We can avoid calling `get_settings_defaults` (just for iterating its keys) by extracting defaults into a static const.
There can be also be another one having an array of keys.

https://github.com/borisbu/OctoRelay/blob/7a5dc76eb65489a3329870f22ff745c18dfb4a53/octoprint_octorelay/__init__.py#L31-L32

Having it as a method makes it create a new object each time while its not needed

RobinTail:
Fixes #103 

- Extracting constants
- Enforcing constraints:
  - on the names of API command
  - on the name of `@` command
  - on repository props
- Simplifying the implementation and probably improving the performance by avoiding repeats of the same operations 
- Cleaner look of the main file
- Moving GPIO setup into constructor (it's called anyway for the `__plugin_implementation__`)


octoprint_octorelay/__init__.py
```diff
@@ -174,9 +55,9 @@ def on_after_startup(self):
 
         self._logger.info("--------------------------------------------")
         self._logger.info("start OctoRelay")
-        settings = self.get_settings_defaults()
+        settings = { **defaultSettings } # clone
```
RobinTail, line 58: this variable is actually redundant, since we don't use it as a whole, only its index within the iteration

octoprint_octorelay/__init__.py
```diff
@@ -385,8 +266,8 @@ def turn_on_pin(self, relay_pin, inverted, cmdON):
         self._logger.info("pin: {} turned on".format(relay_pin))
 
     def update_ui(self):
-        settings = self.get_settings_defaults()
-        for index in settings:
+        settings = { **defaultSettings } # clone
```
RobinTail, line 269: same thing here

octoprint_octorelay/__init__.py
```diff
@@ -5,18 +5,19 @@
...
+from const import defaultSettings, relayIndexes, templates, assets
+from const import permissions, updateConfig, pollingInterval
+from const import apiCommandUpdate, apiCommandGetStatus, apiCommandListAll, atCommand
+
 import flask
 import RPi.GPIO as GPIO
 import os
 
 GPIO.setmode(GPIO.BCM)
 GPIO.setwarnings(False)
```
RobinTail, line 19: perhaps this should be inside the constructor

RobinTail: done

octoprint_octorelay/const.py
RobinTail, line 1: I found that conventional naming for constants is capitalised SNAKE_CASE

RobinTail: fixed

RobinTail (commented):
I like it now

RobinTail:
QA failed ❌ 

```
2023-07-27 21:31:51,780 - octoprint.plugin.core - ERROR - Error loading plugin octorelay
Traceback (most recent call last):
  File "/home/pi/oprint/lib/python3.9/site-packages/octoprint/plugin/core.py", line 1294, in _import_plugin
    module = _load_module(module_name, spec)
  File "/home/pi/oprint/lib/python3.9/site-packages/octoprint/plugin/core.py", line 52, in _load_module
    return imp.load_module(name, f, filename, details)
  File "/home/pi/oprint/lib/python3.9/site-packages/octoprint/vendor/imp.py", line 238, in load_module
    return load_package(name, filename)
  File "/home/pi/oprint/lib/python3.9/site-packages/octoprint/vendor/imp.py", line 212, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 711, in _load
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/pi/oprint/lib/python3.9/site-packages/octoprint_octorelay/__init__.py", line 10, in <module>
    from const import DEFAULT_SETTINGS, RELAY_INDEXES, TEMPLATES, ASSETS
ModuleNotFoundError: No module named 'const'
```

PR #117, opened 2023-07-27
Ref: extracting constants

RobinTail:
Closes #103 

Continuation of #112 — QA failed, wrong import

PR #111, opened 2023-07-27
Ref: replacing `dict()` with curly braces.

RobinTail:
Fixes #107

PR #118, opened 2023-07-28
Releasing 2.2.2

### LINKED ISSUE #103: Default settings should be extracted into static const

### LINKED ISSUE #107: `dict()` could be replaced with curly braces
Reasoning:

- https://stackoverflow.com/questions/2853683/what-is-the-preferred-syntax-for-initializing-a-dict-curly-brace-literals-or
- it's easier to read
- it's JSON compatible object I'm used to working on FE

RobinTail:
Closes #107 
Closes #103 

- #111
- ~~#112~~ (reverted)
- #117


RobinTail (commented):
ok

PR #120, opened 2023-07-28
Ensure booleans: initialisations and strict comparison

### LINKED ISSUE #104: Booleans must be ensured
Replacing `==` with `is (not) True/False` is a refactoring due to [best practices](https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true), preventing type casting — I presume the whole module requires global review on applying that practice. However, certain methods and settings extractions should ensure such things as `inverted` to be boolean for that.

_Originally posted by @RobinTail in https://github.com/borisbu/OctoRelay/pull/44#pullrequestreview-1544079597_

RobinTail:
Closes #104 

Enforcing stricter boolean operations and comparisons

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -262,7 +262,7 @@ class OctoRelayPlugin(
             self.model[index]['state'] = GPIO.input(relay_pin)
             self.model[index]['labelText'] = labelText
             self.model[index]['active'] = active
-            if inverted != self.model[index]['state']:
+            if inverted is not self.model[index]['state']:
                 self.model[index]['iconText'] = iconOn
                 self.model[index]['confirmOff'] = confirmOff
             else:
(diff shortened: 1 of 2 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -249,20 +249,20 @@ def update_ui(self):
             settings[index].update(self._settings.get([index]))
 
             labelText = settings[index]["labelText"]
-            active = int(settings[index]["active"])
+            active = int(settings[index]["active"]) # todo make it bool later
```
RobinTail, line 252: `model::active` should be made boolean later

RobinTail: made #122

RobinTail (commented):
ok, I'm happy with it this way

RobinTail:
QA passed ✅

PR #127, opened 2023-07-28
Readme: new controls image.

RobinTail:
- The screenshot of the settings is almost similar to the one in configuration section
- I'd like to show in the beginning of Readme what user can actually get in the UI


RobinTail (commented):
ok

PR #129, opened 2023-07-28
Distributable ZIP-package

### LINKED ISSUE #115: Switch to ZIP-package distribution channel

The idea is:

Sample: 

Documentation:

RobinTail:
Closes #115

Screenshot 2023-07-29 at 12 00 19

### Benefits

- Can remove JS folder and stop checking that it's in sync with TS
- Can stop doing `Next version` bumping commits

### Details

The PR makes `plugin_version` to exist only within a distributable package.

Documentation on annotated tags:
https://git-scm.com/book/en/v2/Git-Basics-Tagging

Versioning tool:
https://github.com/jbweston/miniver


RobinTail:
The build works surprisingly well.

RobinTail:
It seems that `miniver` I should install myself

RobinTail:
ok, now CI is working as `unknown` which is expected

> Successfully installed OctoRelay-unknown-g474ea56 RPi.GPIO-0.7.1

RobinTail:
OK. Now got to the actual package building with an expected error:

> setuptools.extern.packaging.version.InvalidVersion: Invalid version: 'unknown-g78441c7'

RobinTail:
At this point, instead of `unknown` I'm getting something really weird

```
Invalid version: 'Release-v1.2.0.dev39-g1f4a92c'
```

RobinTail:
Clarified:

> An "annotated" tag is a name in refs/tags that refers to a tag object. Get the tag name (say foo ) and then do a git cat-file -t foo . If it's an an annotated tag, cat-file will tell you that it's a "tag". If it's a simple tag, cat-file will tell you that it's a "commit".

```sh
[local] ➜  OctoRelay git:(zip-package-release) git cat-file -t Release-v1.2.0
tag
[local] ➜  OctoRelay git:(zip-package-release) git cat-file -t 2.3.0a1
commit
```

So, GitHub creates a lightweight tag when doing a release (at least pre-relese)

RobinTail:
The corresponding discussion:
https://github.com/orgs/community/discussions/4924

RobinTail:
🦄  `Successfully installed OctoRelay-2.3.0a3 RPi.GPIO-0.7.1`

RobinTail:
🚀  Got it finally!

Screenshot 2023-07-29 at 01 45 08

.gitattributes
```diff
@@ -0,0 +1 @@
+octoprint_octorelay/_static_version.py export-subst
```
RobinTail, line 1:
```suggestion
octoprint_octorelay/_static_version.py export-subst

```
add EOL

setup.py
```diff
@@ -1,10 +1,27 @@
...
+    return module.__version__, module.get_cmdclass(pkg_path)
+
+miniver_version, cmdclass = get_version_and_cmdclass(r"octoprint_octorelay")
+
 ########################################################################################################################
 # Do not forget to adjust the following variables to your own plugin.
 
 # The plugin's identifier, has to be unique
-from setuptools import setup
+
```
RobinTail, line 24:
```suggestion
```

RobinTail:
✅ workflow operational

RobinTail (commented):
OK. I'm happy with it at this point.
I hope it will work.

RobinTail:
After merging to `master`, the installation URL has to be changed in the plugins listing repo:
https://github.com/OctoPrint/plugins.octoprint.org/blob/gh-pages/_plugins/octorelay.md

RobinTail:
⚠️ QA passed with warnings

Since currently distributed installations refer to archive of the repo sources, installing the next release led to the usual download without JS (no control buttons in UI) and without version set, having weird `unknown` version in Software Update section:

Screenshot 2023-07-30 at 20 01 14

However, LUCKILY, this is great that it's immediately offers to update again. Because now it knows that the new distribution URL is the ZIP-package asset. So, the second update makes it work as expected.

Anyway, the lack of control buttons is misleading and can be disappointing.
We need a stub JS in the sources, notifying users to go to Software Update section and perform the plugin update once again.

PR #131, opened 2023-07-29
Hotfix: Reverting a const for templates.

RobinTail:
Just noticed that in some cases the values of the plugin settings may not appear in OctoPrint.

I have no idea why, but it figured out that keeping templates as inline return fixes this issue.


RobinTail (commented):
OK

PR #132, opened 2023-07-30
JS: notification stub in sources

### LINKED ISSUE #115: Switch to ZIP-package distribution channel

RobinTail:
Closes #115 
Continuation of #129 — QA revealed a need for a stub

⚠️ Breaking
Switching to ZIP-package distribution requires user to update the plugin TWICE.
After the first update there will be no control buttons in navbar which I consider a disruption of service, requiring a user's action to fix.

In order to minimise the confusion I'm making a stub showing a warning icon for attracting user's attention on the needed action, and a dialog explaining what to do.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,9 @@ jobs:
     - name: Install UI dependencies
       working-directory: ts
       run: yarn install
+    - name: Explicitly remove stub
+      working-directory: octoprint_octorelay/static/js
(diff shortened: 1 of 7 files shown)


RobinTail:
Perhaps this should be marked as breaking.

RobinTail (commented):
Should be fine

RobinTail:
✅ QA passed

PR #133, opened 2023-07-30
🚀 Releasing 3.0.0

### LINKED ISSUE #115: Switch to ZIP-package distribution channel

RobinTail:
Closes #115 

- #129 
- #132 

⚠️ The `archive` URL must be changed in the plugin repo of OctoPrint:
https://github.com/OctoPrint/plugins.octoprint.org/blame/ee9e29494865fb989ef707473f010939f692579d/_plugins/octorelay.md#L17


RobinTail:
The latest release asset URL was wrong. The correct one is tested based on this article:

https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release

Fixed in 6ee1694

RobinTail (commented):
OK. Let's do another RC before merging.

RobinTail:
Let's try one more to ensure no `-dirty` suffix

RobinTail:
✅ QA passed

RobinTail:
The corresponding PR for updating the `archive` URL in plugins repo:
https://github.com/OctoPrint/plugins.octoprint.org/pull/1219

PR #136, opened 2023-07-31
Revert: const for default settings.

RobinTail:
Again.
It turned out that for some weird reason, keeping default settings in a dedicated const, even when making its copy in method, leads to a bug — settings are being saved partially (only those that were changed, while the rest is being reset to defaults). There is some magic happening there that I don't understand.


octoprint_octorelay/__init__.py
```diff
@@ -40,7 +40,136 @@ def __init__(self):
             self.model[index] = {}
 
     def get_settings_defaults(self):
-        return DEFAULT_SETTINGS
```
RobinTail, line 43: Dear @jneilliii , do you know why this does not work well?
And even this:

```py
 def get_settings_defaults(self):
        return DEFAULT_SETTINGS.copy()
```

why it makes setting saving partial?
Why do I have to keep static object inside the method to make it work?
I almost broke my mind on it.

RobinTail:
✅ QA passed

octoprint_octorelay/__init__.py
RobinTail: I think I figured this out.
`.copy()` keeps the references to the nested objects, even they are not assigned to any const/vars.

Screenshot 2023-08-01 at 01 11 32

PR #137, opened 2023-07-31
Immutable: Replacing const with getters

RobinTail:
Once I figured out the issue with mutability and copies, I realized that I can still do the desired arrangement, but making a getter fn in a separate file

https://github.com/borisbu/OctoRelay/pull/136#discussion_r1279968183


RobinTail:
nice! 🦄

PR #138, opened 2023-08-01
DNRY: Relay driver

### LINKED ISSUE #114: DNRY: Should reuse `turn_on/off_pin` methods
Avoid repeating same commands in different places of the implementation by reusing the existing methods for toggling relays

https://github.com/borisbu/OctoRelay/blob/c7f2ffa64b2e6bfb5717388cfb04cd116c295ff3/octoprint_octorelay/__init__.py#L190-L193

RobinTail:
Closes #114 

Several statements in the original code are being repeated multiple times:

- Setting and unsetting warnings
- Setting pin output mode
- Coverting pin state to relay state ("XOR with inverted")

This can lead to mistakes in the future and complicated the further development.
I'd like to extract the implementation of operating relays into a driver, responsible for knowledge on how to deal with relays having pins and properties like "inverted"

This PR also makes the plugin class only refer to the RELAY state (not pin state).


octoprint_octorelay/test_driver.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+import unittest
+import sys
+from unittest.mock import Mock
+
+GPIO_mock = sys.modules["RPi.GPIO"] # mocked in test__init.py
```
RobinTail, line 6: in both test files GPIO mocking should be make conditional, so each file could be ran individually as well

octoprint_octorelay/__init__.py
```diff
@@ -32,7 +33,6 @@ class OctoRelayPlugin(
     def __init__(self):
         # pylint: disable=super-init-not-called
         GPIO.setmode(GPIO.BCM)
```
RobinTail, line 35: should this also be a part of the driver?

octoprint_octorelay/__init__.py
```diff
@@ -139,29 +134,15 @@ def on_api_command(self, command, data):
...
             cmd_off = settings["cmdOFF"]
+            self._logger.debug(f"OctoRelay before update {relay}")
 
-            GPIO.setwarnings(False)
-
-            GPIO.setup(relay_pin, GPIO.OUT)
-            # XOR with inverted
-            relay_state = inverted is not bool(GPIO.input(relay_pin))
-
-            self._logger.debug(f"OctoRelay before pin: {relay_pin}, inverted: {inverted}, relay_state: {relay_state}")
```
RobinTail, line 154: there was logging of before-state, missing in the serialization

octoprint_octorelay/__init__.py
```diff
@@ -227,46 +208,37 @@ def print_stopped(self):
...
+            )
             # set the icon state
-            self.model[index]["relay_pin"] = relay_pin
-            self.model[index]["state"] = pin_state # int
+            self.model[index]["relay_pin"] = relay.pin
+            self.model[index]["state"] = int(relay.get_pin_state()) # int
             self.model[index]["labelText"] = settings[index]["labelText"]
             self.model[index]["active"] = int(settings[index]["active"]) # todo make it bool later
-            if relay_state:
+            if relay.is_closed(): # todo perhaps needs a cache
```
RobinTail, line 241: can I avoid actual reading of the pin state twice?

octoprint_octorelay/__init__.py
```diff
@@ -293,14 +265,17 @@ def get_update_information(self):
...
         self._logger.debug("input_polling")
         for index in RELAY_INDEXES:
             # model::active is currently int, see update_ui()
             active = bool(self.model[index]["active"])
-            model_state = self.model[index]["state"] # int
-            actual_state = GPIO.input(self.model[index]["relay_pin"]) # int
+            model_state = bool(self.model[index]["state"]) # todo make state bool later
+            actual_state = Relay(
+                int(self.model[index]["relay_pin"]),
+                False # does not matter for pin state
```
RobinTail, line 277: perhaps it should be made optional in this case

octoprint_octorelay/driver.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+from typing import Optional
+from RPi import GPIO
+
+def xor(left: bool, right: bool) -> bool:
+    return left is not right
+
+class Relay():
+    def __init__(self, pin: int, inverted: bool):
```
RobinTail, line 9: should I use an object or named arguments to support more props in the future?

octoprint_octorelay/test__init.py
```diff
@@ -10,10 +10,14 @@
...
-GPIO_mock.BCM = "MockedBCM"
-GPIO_mock.OUT = "MockedOUT"
-sys.modules["RPi.GPIO"] = GPIO_mock
+if "RPi.GPIO" in sys.modules:
+    GPIO_mock = sys.modules["RPi.GPIO"]
+else:
+    GPIO_mock = Mock()
+    GPIO_mock.BCM = "MockedBCM"
+    GPIO_mock.OUT = "MockedOUT"
+    sys.modules["RPi.GPIO"] = GPIO_mock
```
RobinTail, line 19: perhaps we can actually replace GPIO assertions to call-assertions of the driver,
since there are reliable driver tests already.

RobinTail: attempted to do this, but it does not look good enough so far

octoprint_octorelay/__init__.py
RobinTail: fixed

octoprint_octorelay/__init__.py
RobinTail: fixed, also featuring `pin_state` (bool) instead of `state` (int)

octoprint_octorelay/__init__.py
RobinTail: done

octoprint_octorelay/driver.py
RobinTail: so far I'm fine with it as it is

octoprint_octorelay/__init__.py
```diff
@@ -293,15 +259,18 @@ def get_update_information(self):
...
+    # Polling thread
     def input_polling(self):
         self._logger.debug("input_polling")
         for index in RELAY_INDEXES:
             # model::active is currently int, see update_ui()
             active = bool(self.model[index]["active"])
-            model_state = self.model[index]["state"] # int
-            actual_state = GPIO.input(self.model[index]["relay_pin"]) # int
-            if active and actual_state != model_state:
+            model_state = self.model[index]["pin_state"] # bool since v3.1
```
RobinTail, line 268: can I switch to relay state anyway?

octoprint_octorelay/__init__.py
```diff
@@ -227,46 +199,40 @@ def print_stopped(self):
...
+            pin_state = relay.get_pin_state()
             # set the icon state
-            self.model[index]["relay_pin"] = relay_pin
-            self.model[index]["state"] = pin_state # int
+            self.model[index]["relay_pin"] = relay.pin
+            self.model[index]["pin_state"] = pin_state # bool since v3.1
             self.model[index]["labelText"] = settings[index]["labelText"]
             self.model[index]["active"] = int(settings[index]["active"]) # todo make it bool later
-            if relay_state:
+            if relay.get_state_from_pin_state(pin_state):
```
RobinTail, line 235: It would be great to get rid of this in order to make an abstract driver later

RobinTail: done

octoprint_octorelay/driver.py
```diff
@@ -0,0 +1,47 @@
...
+from RPi import GPIO
+
+# The driver operated BCM mode of pins enumeration
+GPIO.setmode(GPIO.BCM)
+
+def xor(left: bool, right: bool) -> bool:
+    return left is not right
+
+class Relay():
+    def __init__(self, pin: int, inverted: Optional[bool]):
```
RobinTail, line 12:
```suggestion
    def __init__(self, pin: int, inverted: bool):
```

PR #139, opened 2023-08-02
Moving the tests into a dedicated folder.

RobinTail:
This will also help to reduce the distributable release-file by excluding test files from it.


RobinTail (commented):
ok

PR #147, opened 2023-08-03
Consistent naming for the plugin settings.

RobinTail:
- [x] New naming
- [x] migration
- [x] Update the bindings in settings template
- [x] A test for bindings
- ~~I also would like to add settings [preprocessors]~~(https://docs.octoprint.org/en/master/plugins/mixins.html#octoprint.plugin.SettingsPlugin.get_settings_preprocessors)
  - no, this is not needed, at least for now


RobinTail (commented):
ok

PR #148, opened 2023-08-03
Relative imports

RobinTail:
I tried before, but it didn't work when tests were in the same directory with the plugin module.


RobinTail:
should be fine, tried this locally

Issue #28, opened 2021-06-12
Feature: Auto ON after print


MelGrubb:
I'd like my cooldown fan to turn on for X minutes when a print is completed. There is currently an option to turn relays OFF after a print, but no way to turn them ON.

Sebbabas:
hello,
Yes this option would be useful for me too.

mblacklin1:
Agreed - would love to see this implemented somehow.

RobinTail:
I wanna change settings interface for this, providing a sort of matrix

| Event | Auto ON | Auto OFF |
|--------|:------:|:------:|
| Initially |☑️|  ☑️  |
| Before print | ☑️  | ☑️  |
| After print | ☑️  | ☑️  |

RobinTail:
@MelGrubb @PalmyreSG1 @Sebbabas @mblacklin1 ,

The feature is available in the recently released version 3.4.0 🚀

RobinTail:
>  fan to turn on for X minutes

The feature on turning something on for a limited time released in v3.8.0 🚀 
@MelGrubb

PR #166, opened 2023-08-10
Feat: duration being turned ON (turn off after turned on)

### LINKED ISSUE #27: Feature: auto OFF not related to printing
Thank you
Great plugin
Is it possible to add the timed shutdown function without starting prints?
It would be very convenient for me to manage the buttons of the Sunlu Dryer Box S1, it starts up via relay but then you have to give 1 click to make it turn on and bring it to temperature, so if I had the possibility to give 1-3 seconds to the relay I could connect it to the button physical to start it remotely.
Did I have a piece of code to add?

Thank you

RobinTail:
An additional feature for #28 .
Closes #27 

By setting:
- on Printing Stopped: ON
- after Turned ON: OFF, delay X

User can set the limit for the relay being turned ON.

Screenshot 2023-08-16 at 18 42 45


octoprint_octorelay/const.py
```diff
@@ -5,13 +5,14 @@
...
 PRINTING_STOPPED = "PRINTING_STOPPED"
+TURNED_ON = "TURNED_ON"
 
 # Task cancellation exceptions
 # { eventHappened: [ events which postponed timers should NOT be cancelled ]
 CANCELLATION_EXCEPTIONS = {}
 
 # Versioning of the plugin's default settings described below
-SETTINGS_VERSION = 3
+SETTINGS_VERSION = 4
```
RobinTail, line 15: I'm not sure if migration is needed in this case

RobinTail, line 15:
```suggestion
SETTINGS_VERSION = 3
```

I'm not sure if migration is needed in this case

octoprint_octorelay/__init__.py
```diff
@@ -196,8 +196,11 @@ def toggle_relay(self, index, target: Optional[bool] = None):
             f"Toggling relay {index} on pin {pin}" if target is None else
             f"Turning the relay {index} {'ON' if target else 'OFF'} (pin {pin})"
         )
-        cmd = settings["cmd_on" if relay.toggle(target) else "cmd_off"]
+        state = relay.toggle(target)
+        cmd = settings["cmd_on" if state else "cmd_off"]
         self.run_system_command(cmd)
+        if state:
+            self.handle_plugin_event(TURNED_ON)
```
RobinTail, line 203: - [ ] here is the problem: THIS kind of event is only related to the particular relay, it has a scope, or payload, which should be handled by `handle_plugin_event`

octoprint_octorelay/__init__.py
```diff
@@ -196,8 +196,11 @@ def toggle_relay(self, index, target: Optional[bool] = None):
             f"Toggling relay {index} on pin {pin}" if target is None else
             f"Turning the relay {index} {'ON' if target else 'OFF'} (pin {pin})"
         )
-        cmd = settings["cmd_on" if relay.toggle(target) else "cmd_off"]
+        state = relay.toggle(target)
+        cmd = settings["cmd_on" if state else "cmd_off"]
         self.run_system_command(cmd)
+        if state:
+            self.handle_plugin_event(TURNED_ON, scope = [index])
```
RobinTail, line 203: - [ ] here is another potential issue: infinite loops: should prevent `relay.state=target` (same state) toggling or it's side effects

RobinTail: OR! I can forbid to event TURNED_ON having action ON

octoprint_octorelay/__init__.py
```diff
@@ -168,9 +168,9 @@ def on_event(self, event, payload):
         #elif event == Events.PRINT_CANCELLED:
             # self.print_stopped()
 
-    def handle_plugin_event(self, event):
+    def handle_plugin_event(self, event, scope = RELAY_INDEXES):
         settings = self._settings.get([], merged=True) # expensive
-        for index in RELAY_INDEXES:
+        for index in scope:
```
RobinTail, line 173: `cancel_tasks` should also comply the scope in this case

tests/snapshots/snap_test_templates.py
```diff
@@ -532,6 +532,59 @@
...
+                        <div class="btn-group">
+                            
+                            <!--ko let: { classBinding: {
+                                'active btn-success': state() === true,
+                                'btn-default': state() !== true
+                            } } -->
+                            <label class="btn" data-bind="css: classBinding">
+                                <input
+                                    type="radio"
+                                    data-bind="checkedValue: true, checked: state"
```
RobinTail, line 547: ON button should be disabled for this event

octoprint_octorelay/__init__.py
```diff
@@ -178,14 +178,16 @@ def on_event(self, event, payload):
...
+        self._logger.debug(f"Handling the plugin event {event} having scope: {scope}")
         settings = self._settings.get([], merged=True) # expensive
-        for index in RELAY_INDEXES:
+        for index in scope:
             if bool(settings[index]["active"]):
                 target = settings[index]["rules"][event]["state"]
                 if target is not None:
                     target = bool(target)
+                    if target and event == TURNED_ON:
+                        return # avoid infinite loop
```
RobinTail, line 190: better would be not `return`, but something like `continue` — proceed with the next relay.

```suggestion
                        return # avoid infinite loop
```

octoprint_octorelay/__init__.py
```diff
@@ -178,14 +178,16 @@ def on_event(self, event, payload):
...
+        self._logger.debug(f"Handling the plugin event {event} having scope: {scope}")
         settings = self._settings.get([], merged=True) # expensive
-        for index in RELAY_INDEXES:
+        for index in scope:
             if bool(settings[index]["active"]):
                 target = settings[index]["rules"][event]["state"]
                 if target is not None:
                     target = bool(target)
+                    if target and event == TURNED_ON:
+                        continue # avoid infinite loop
```
RobinTail, line 190:
```suggestion
                    if target and event == TURNED_ON:
                        self._logger.debug(f"Skipping {index} to avoid infinite loop")
                        continue # avoid infinite loop
```

RobinTail (commented):
ok

PR #171, opened 2023-08-12
Feat: Ability to cancel postponed toggling

RobinTail:
Closes #87 

### Dialog

Screenshot 2023-08-14 at 19 23 27

### New setting (opt-out)

Screenshot 2023-08-14 at 14 28 23

### Tasks

- [x] tests
- [x] countdown: 1 minute / 1 second
- [x] disposer
- [x] opt-in/out these notifications

PR #173, opened 2023-08-14
🚀 3.7.0

### LINKED ISSUE #87: (Feature Request)  Off after printing delay cancel button
Capture

First of all I have to say, I love this plugin, I truly use it everyday, all the time. Thank you for maintaining it.  I have my setup on a relay controlled power strip, that as you can see in the picture, after the print is finished, basically turns everything off after 300 seconds, (5 Minutes).   I can just leave a print going, and when its done everything is turned off. It works amazing, I don't have to worry about it all being left on, after a long print that I may leave running as I go to work or whatever. When its done printing, after 5 minutes of fan cool down it all shuts off.  What I would like to see, is a cancel button for that timer, to cancel the process so it does not turn the relay off. As it stands, since I have it set for 5 minutes, if I don't have another print ready to go, and start printing within that time, It shuts the relay off, which in turn shuts my printer, lights, fans and everything off (which it is supposed to). I  then have to manually turn it back on and wait for it to reboot my printer, and the reconnect with octoprint. Its not terrible, but it gets annoying after you have to do it a few hundred times.  If there was maybe a little button that just says cancel, that pops up next to the printer on and off icon, or whatever relay your particular setup is using, that would allow to cancel the "OFF after printing delay" That would be amazing. Or if it maybe had a countdown time with the seconds left before it shuts the relay off in the button so you know how much time is left to cancel it. That would be awesome. I don't know enough python to attempt to try and do it myself.

Sorry if I didn't post this in the right place, this is my first time ever posting on here, I actually just signed up just to ask this. 

Again, thank you for maintaining this awesome plugin, and keep up all the good work.

RobinTail:
Closes #87 

- #171


RobinTail:
✅  QA passed, but found a couple unrelated issues:

- The OS command can affect the performance, perhaps it should not be awaited
- Non-active relays should not be requested for the pin state in `update_ui`

PR #176, opened 2023-08-14
Avoid requesting disabled relay state for UI

### LINKED ISSUE #175: Disabled relays should not request the pin state
- Non-active relays should not be requested for the pin state in `update_ui`

_Originally posted by @RobinTail in https://github.com/borisbu/OctoRelay/issues/173#issuecomment-1677837802_

RobinTail:
Closes #175 

- Also fixing invalid mock resetting method used in tests

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -239,7 +239,7 @@ class OctoRelayPlugin(
     def update_ui(self):
         settings = self._settings.get([], merged=True) # expensive
         upcoming = self.get_upcoming_tasks(filter(
-            lambda index: bool(settings[index]["active"]) and settings[index]["show_upcoming"],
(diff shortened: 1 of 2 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -239,21 +239,22 @@ def reducer(agg, task):
     def update_ui(self):
         settings = self._settings.get([], merged=True) # expensive
         upcoming = self.get_upcoming_tasks(filter(
-            lambda index: settings[index]["show_upcoming"],
+            lambda index: bool(settings[index]["active"]) and settings[index]["show_upcoming"],
```
RobinTail, line 242: they both should be `bool()`

RobinTail:
```suggestion
            lambda index: bool(settings[index]["active"]) and bool(settings[index]["show_upcoming"]),
```

should still fit

RobinTail (commented):
ok

PR #186, opened 2023-08-16
Introducing UI update lock.

### LINKED ISSUE #184: UI update can be triggered twice
the UI update can be triggered twice (by action and by polling).
It needs a lock.

```
2023-08-16 23:05:21,643 - octoprint.plugins.octorelay - INFO - Received the API command update with parameters: {'pin': 'r3', 'command': 'update'}
2023-08-16 23:05:21,643 - octoprint.plugins.octorelay - DEBUG - Requested to switch the relay r3
2023-08-16 23:05:21,759 - octoprint.plugins.octorelay - DEBUG - Toggling the relay r3 on pin 18
2023-08-16 23:05:21,760 - octoprint.plugins.octorelay - DEBUG - Updating the UI
2023-08-16 23:05:21,799 - octoprint.plugins.octorelay - DEBUG - relay: r3 has changed its pin state
2023-08-16 23:05:21,799 - octoprint.plugins.octorelay - DEBUG - Updating the UI
```

improved logging helped to reveal it.

_Originally posted by @RobinTail in https://github.com/borisbu/OctoRelay/issues/183#issuecomment-1681275468_

RobinTail:
Closes #184


octoprint_octorelay/__init__.py
```diff
@@ -262,6 +263,9 @@ def reducer(agg, task):
         )
 
     def update_ui(self):
+        if self.ui_update_lock:
+            return
```
RobinTail, line 267: i don't like this approach, since the next call of this method can be caused by the most recent data state, while the payload for ongoing execution may already have been shaped.
that can lead to missing the important updates.
what is really required is an actual thread lock that awaits to be released.

RobinTail:
> i don't like this approach, since the next call of this method can be caused by the most recent data state, while the payload for ongoing execution may already have been shaped. that can lead to missing the important updates. what is really required is an actual thread lock that awaits to be released.

yes, but this way it does not solve the original issue

RobinTail:
Hey, girls, I have an idea — we should not call the UI update fn in the polling thread during the ongoing one caused by the main thread.
The simple bool should be enough.

RobinTail:
Looks better.
It's always nice to talk to a smart person, thanks for stepping in.

octoprint_octorelay/__init__.py
```diff
@@ -315,6 +318,8 @@ def get_update_information(self):
     # Polling thread
     def input_polling(self):
         # self._logger.debug("input_polling") # in case your log file is too small
+        if self.ui_update_lock:
+            return # issue 186, avoid update during making another update
```
RobinTail, line 322:
```suggestion
            return # issue 186, avoid the update during the another one
```

RobinTail:
Now we can call the UI update fn wherever it's needed, without relying on the polling thread.
For instance — the relay toggling fn.

RobinTail (commented):
ok

PR #188, opened 2023-08-17
Improving documentation: the plugin indicates the relay state too

### LINKED ISSUE #187: Indicating the relay state
There is another plugin called PSU Control that has a sensing option to monitor the GPIO pins on the Pi.
It gives a real indication of the relay status as opposed to just an On/Off toggle.

Do you think this would be a good feature for OctoRelay too?

image

image

RobinTail:
Closes #187 

This also reflects the recent changes I made in #166


RobinTail (commented):
ok

PR #192, opened 2023-08-18
Removing the outside click listener for upcoming switch popover.

### LINKED ISSUE #190: UI update does not happen in event of `PRINTING_STOPPED`
I have been gone for a few days with work, and came back to see this update :)   I was excited to try it out, and unfortunately it did not work for me at first, and I am not sure its working properly for me still.   I set up a short print, and when it finished, nothing happened at all.  I rebooted everything and tried again, Still nothing happened. I tried hovering my mouse over the icon and still nothing.  I then tried disabling that option and clicked the save.  then re-enabled it, and clicked save. Then the notification pops up. It only works if you disable it and re-enable it within the timeout period.  I then tried leaving it enabled and then refreshing the browser page, and that worked also. It does not however automatically pop up for me, unless I do those particular things. I an assuming it should automatically pop up without a manual browser refresh.   I have uploaded some pictures of my octorelay version. Also my configuration. I am using Themify with the dark theme, but I don't know if that would cause the problem though.  I have included pictures of all versions.  
RelaySettings
version
Theme
header

_Originally posted by @backupartist in https://github.com/borisbu/OctoRelay/issues/87#issuecomment-1679954064_

RobinTail:
Probably fixes #190 

This is the UX adjustment that removes the outside click listener from the popover, that might lead to confusion, making the popover disappear unexpectedly from the user's perspecive.

### DIFF
--- a/ts/octorelay.spec.ts
+++ b/ts/octorelay.spec.ts
@@ -246,35 +246,6 @@ describe("OctoRelayViewModel", () => {
     }
   );
 
-  test("Should display upcoming state popover (%s sec delay)", () => {
-    const handler = (registry[0].construct as OwnModel & OwnProperties)
-      .onDataUpdaterPluginMessage;
-    handler("octorelay", {
-      r1: {
-        relay_pin: 16,
-        inverted_output: false,
-        relay_state: true,
-        label_text: "Nozzle Light",
-        active: true,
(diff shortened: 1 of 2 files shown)


RobinTail:
I'm going to try testing this in a RC version distributed via Prerelease channel as a possible cause for the issue.

PR #207, opened 2023-09-15
Disconnecting from the printer before when turning its relay off

### LINKED ISSUE #153: Feature: disconnect before turning printer relay OFF
> In a same way autodisconnect would be great. Disconnect printer before relay is deactivated would avoid "SerialException" error from octoprint.

This would require for OctoRelay to know which relay controls the printer.

_Originally posted by @RobinTail in https://github.com/borisbu/OctoRelay/issues/18#issuecomment-1664642386_

RobinTail:
Closes #153 

This feature makes the following improvements:
- it prevents the `SerialException` error appearance when turning the printer relay off,
- it prevents the redundant attempts to connect when there are no printers visible.

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -171,8 +171,9 @@ class OctoRelayPlugin(
             self.handle_plugin_event(PRINTING_STOPPED)
         elif hasattr(Events, "CONNECTIONS_AUTOREFRESHED"): # Requires OctoPrint 1.9+
             if event == Events.CONNECTIONS_AUTOREFRESHED:
-                self._logger.debug("Connecting to the printer")
-                self._printer.connect()
+                if payload is not None and "ports" in payload and len(payload["ports"]) > 0:
+                    self._logger.debug("Connecting to the printer")
+                    self._printer.connect()
         #elif event == Events.PRINT_CANCELLING:
             # self.print_stopped()
         #elif event == Events.PRINT_CANCELLED:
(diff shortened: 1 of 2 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -171,8 +171,9 @@ def on_event(self, event, payload):
             self.handle_plugin_event(PRINTING_STOPPED)
         elif hasattr(Events, "CONNECTIONS_AUTOREFRESHED"): # Requires OctoPrint 1.9+
             if event == Events.CONNECTIONS_AUTOREFRESHED:
-                self._logger.debug("Connecting to the printer")
-                self._printer.connect()
+                if payload is not None and "ports" in payload and len(payload["ports"]) > 0:
```
RobinTail, line 174: this prevents:
- redundant attempts to connect when this event being fired initially without ports
- attempts to connect after disconnecting (the event is also fired)

octoprint_octorelay/__init__.py
```diff
@@ -209,11 +210,20 @@ def handle_plugin_event(self, event, scope = None):
...
+
     def toggle_relay(self, index, target: Optional[bool] = None):
         settings = self._settings.get([index], merged=True) # expensive
         if not bool(settings["active"]):
             self._logger.debug(f"Refusing to switch the relay {index} since it's disabled")
             return
+        if self.is_printer_relay(index):
+            self._logger.debug(f"{index} is the printer relay")
+            if self._printer.is_operational():
+                self._logger.debug(f"Disconnecting from the printer before turning {index} OFF")
```
RobinTail, line 225: the `target` state can actually be `None` meaning the toggling.
the relay's current state is unknown at this point as well.
therefore I'm intentionally referring to the printer's state, which makes more sense for the action of disconnecting, so the target relay state is assumed from the printer's state here.

tests/test_init.py
```diff
@@ -505,9 +505,21 @@ def test_update_ui(self):
                 "MockedIdentifier", expected_model
             )
 
+    def test_is_printer_relay(self):
```
RobinTail, line 508:
```suggestion
    def test_is_printer_relay(self):
        # Should assert the equality of the common/printer value with the supplied argument
```

RobinTail (commented):
ok

octoprint_octorelay/__init__.py
RobinTail: However, I should exclude the case when the `target` is `True`.
Can be some postponed task which leads to turning the relay even when it's already on. ⚠️

RobinTail: done in 46e75f0

RobinTail (commented):
now ok

PR #209, opened 2023-09-18
🚀 3.9.1

### LINKED ISSUE #205: Auto-off timer is not canceled with a new print

RobinTail:
Accumulating PR

Fixes #205 

- #208
- #213 
- #214


RobinTail:
🚀 3.9.1rc1

RobinTail:
☑️ The issue fixed, but QA did not pass:

- The UI should be updated when any of the postponed tasks cancelled even if no others actions taken

RobinTail:
🚀 3.9.1rc2

RobinTail:
✅ QA passed

PR #213, opened 2023-09-19
Fix task cancellation when `on Printing Started` is `skip`

### LINKED ISSUE #205: Auto-off timer is not canceled with a new print

RobinTail:
Fixes #205


RobinTail:
I think that `TURNED_ON` event should be an exception. A low priority event that should not cancel the startup and printing event based tasks.

octoprint_octorelay/const.py
```diff
@@ -8,9 +8,19 @@
...
+# Highest priority is 1.
+PRIORITIES = {
+    USER_ACTION: 1,
+    STARTUP: 2,
+    PRINTING_STARTED: 2,
+    PRINTING_STOPPED: 2,
+    TURNED_ON: 3
+}
+
+# Missing events above will be treated as having this priority
```
RobinTail, line 21:
```suggestion
# Missing events above will be treated as ones having this priority
```

RobinTail (commented):
ok.
if the exceptional state of TURNED_ON event will have questionable side effects, it will be easy to fix later

PR #214, opened 2023-09-20
Ensure UI update when event just cancels tasks

### LINKED ISSUE #205: Auto-off timer is not canceled with a new print

RobinTail:
Closes #205 
Addition to #213 
Revealed during the QA in #209


RobinTail (commented):
ok

PR #215, opened 2023-09-20
Remove `typing-extensions`

### LINKED ISSUE #212: Task: Drop Python 3.7 (EOL)
Its support should be dropped.

https://devguide.python.org/versions/

RobinTail:
Closes #212 and removes `typing-extensions` dependency


RobinTail:
Now can improve typing
https://docs.python.org/3/whatsnew/3.8.html#typing

RobinTail:
Consider also to review:
- https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions
- https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

octoprint_octorelay/__init__.py
```diff
@@ -38,7 +38,7 @@ class OctoRelayPlugin(
     def __init__(self):
         # pylint: disable=super-init-not-called
         self.polling_timer = None
-        self.tasks = [] # of Task
+        self.tasks: List[Task] = []
```
RobinTail, line 41: From 3.9 can be just `list[Task]`, but in 3.8 have to use `typing`

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#useful-built-in-types

RobinTail (commented):
ok

PR #217, opened 2023-09-21
AutoConnect delay

### LINKED ISSUE #189: AutoConnect delay
I am using a relay switched power strip that physically turns everything off.  I love this feature, however it does not work properly for me. Since I use that power strip and its having to completely reboot, it times out before my printer is fully booted up in order to be detected and connected, Is there a delay setting for this, that I can set before it tries to AutoConnect? That would give my printer time to fully boot up.

_Originally posted by @backupartist in https://github.com/borisbu/OctoRelay/issues/18#issuecomment-1679966520_

RobinTail:
Closes #189 

image


RobinTail (commented):
Ok. Need to try this in action.

RobinTail:
✅ QA passed

PR #233, opened 2023-10-05
Supporting Python 3.12

RobinTail:


RobinTail:
Should work on this when OctoPrint releases its own support of 3.12, which is currently in `devel` branch only.

PR #238, opened 2023-10-16
Readme: Fix quoted illustrations

RobinTail:
Something changed in GitHub markdown, so quoting images does not work anymore.


RobinTail (commented):
ok

PR #240, opened 2023-10-17
Ref: API: Moving response shaping out of handlers

### LINKED ISSUE #164: API inconsistency and payload props naming
This is actually not a pin, it's a relay index (`r1..r8`).

https://github.com/borisbu/OctoRelay/blob/fc54cf0344a1a60827dfbaf1e94b2ec483975533/octoprint_octorelay/__init__.py#L84-L85

It should be renamed.
In order to do this without breaking changes, we should support backward compatibility somehow.

RobinTail:
Contributes to #164 

No breaking changes to the API yet, but preparing for introducing its new version by extracting the response shaping outta handlers. So the `on_api_command` method will be the only one who operates `flask`.

Handlers, on the other hand, are either return data OR raise exceptions.

For that, introducing also `HandlingException` class, accepting the HTTP status code that is supposed to be used for managing the response in case of an error raised during handling the request.

Preserving (for now) the original behaviour of `update` command, which in case of disabled relay responds with `status: error` instead of aborting.

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -121,10 +121,12 @@ class OctoRelayPlugin(
     def handle_get_status_command(self, index: str) -> bool:
         self._logger.debug(f"Getting the relay {index} state")
         settings = self._settings.get([index], merged=True) # expensive
+        if not bool(settings["active"]):
+            raise HandlingException(400)
         return Relay(
             int(settings["relay_pin"] or 0),
             bool(settings["inverted_output"])
-        ).is_closed() if bool(settings["active"]) else False
+        ).is_closed()
 
     def handle_update_command(self, index: str, target: Optional[bool] = None) -> bool:
         self._logger.debug(f"Requested to switch the relay {index} to {target}")
@@ -154,7 +156,10 @@ class OctoRelayPlugin(
             self._logger.info(f"Responding to {LIST_ALL_COMMAND} command: {response}")
             return flask.jsonify(response)
         if command == GET_STATUS_COMMAND: # API command to get relay status
-            is_closed = self.handle_get_status_command(data["pin"])
+            try:
+                is_closed = self.handle_get_status_command(data["pin"])
+            except HandlingException:
+                is_closed = False # todo should just abort in the next version
             self._logger.info(f"Responding to {GET_STATUS_COMMAND} command: {is_closed}")
             return flask.jsonify({"status": is_closed})
         if command == UPDATE_COMMAND: # API command to toggle the relay
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -137,10 +137,11 @@ class OctoRelayPlugin(
         self.update_ui()
         return state
 
-    def handle_cancel_task_command(self, subject: str, target: bool, owner: str):
+    def handle_cancel_task_command(self, subject: str, target: bool, owner: str) -> bool:
         self._logger.debug(f"Cancelling tasks from {owner} to switch the relay {subject} {'ON' if target else 'OFF'}")
-        self.cancel_tasks(subject, USER_ACTION, target, owner)
+        is_cancelled = self.cancel_tasks(subject, USER_ACTION, target, owner)
         self.update_ui()
+        return is_cancelled
 
     def on_api_command(self, command, data):
         self._logger.info(f"Received the API command {command} with parameters: {data}")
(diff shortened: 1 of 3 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -115,54 +116,64 @@ def handle_list_all_command(self):
...
-            return flask.abort(403)
+            raise HandlingException(403)
         if index not in RELAY_INDEXES:
             self._logger.warn(f"Invalid relay index supplied: {index}")
-            return flask.jsonify(status="error")
+            raise HandlingException(400)
         try:
-            state = self.toggle_relay(index, target)
+            state = self.toggle_relay(index, target) # can raise an Exception if relay is disabled
         except Exception as exception:
```
RobinTail, line 141:
```suggestion
            state = self.toggle_relay(index, target)
        except Exception as exception: # disabled relay
```

octoprint_octorelay/__init__.py
```diff
@@ -115,54 +116,64 @@ def handle_list_all_command(self):
...
-        self._logger.debug(f"Responding to {UPDATE_COMMAND} command. Switched state to {state}")
-        return flask.jsonify(status="ok", result=state)
+        return state
 
-    def handle_cancel_task_command(self, subject: str, target: bool, owner: str):
+    def handle_cancel_task_command(self, subject: str, target: bool, owner: str) -> bool:
         self._logger.debug(f"Cancelling tasks from {owner} to switch the relay {subject} {'ON' if target else 'OFF'}")
-        self.cancel_tasks(subject, USER_ACTION, target, owner)
+        is_cancelled = self.cancel_tasks(subject, USER_ACTION, target, owner)
         self.update_ui()
```
RobinTail, line 149: maybe can reduce updating UI only to the cases when cancelling tasks succeeded

tests/test_init.py
```diff
@@ -871,22 +872,36 @@ def test_handle_get_status_command(self, jsonify_mock):
...
+            self.assertEqual(
+                self.plugin_instance.handle_get_status_command("r4"),
+                case["expectedStatus"]
+            )
             self.plugin_instance._settings.get.assert_called_with(["r4"], merged=True)
-            jsonify_mock.assert_called_with(status=case["expectedStatus"])
 
-    @patch("flask.jsonify")
+    def test_handle_get_status_command__exception(self):
+        relay_settings_mock = {
```
RobinTail, line 882:
```suggestion
    def test_handle_get_status_command__exception(self):
        # Should raise an Exception when requesting the state of disabled relay
        relay_settings_mock = {
```

tests/test_init.py
```diff
@@ -935,16 +957,9 @@ def test_handle_update_command(self, system_mock, jsonify_mock):
...
-                    jsonify_mock.assert_called_with(status=case["expectedStatus"], result=case["expectedResult"])
-                else:
-                    jsonify_mock.assert_called_with(status=case["expectedStatus"])
 
-
-    @patch("flask.abort")
-    def test_handle_update_command__exception_permissions(self, abort_mock):
-        # Should refuse to update the relay state in case of insufficient permissions
+    def test_handle_update_command__exception_permissions(self):
+        # Should raise when attempting in case of insufficient permissions
```
RobinTail, line 962:
```suggestion
        # Should raise in case of insufficient permissions
```

tests/test_init.py
```diff
@@ -871,22 +872,37 @@ def test_handle_get_status_command(self, jsonify_mock):
...
+            self.assertEqual(
+                self.plugin_instance.handle_get_status_command("r4"),
+                case["expectedStatus"]
+            )
             self.plugin_instance._settings.get.assert_called_with(["r4"], merged=True)
-            jsonify_mock.assert_called_with(status=case["expectedStatus"])
 
-    @patch("flask.jsonify")
+    def test_handle_get_status_command__exception(self):
+        # Should raise an Exception when requesting the state of disabled relay
```
RobinTail, line 882:
```suggestion
        # Should raise when requesting the state of disabled relay
```

tests/test_init.py
```diff
@@ -968,53 +982,85 @@ def test_handle_update_command__exception_disabled(self, jsonify_mock):
...
+            { "status": 400, "expectedMethod": jsonify_mock, "expectedArgument": {"status": "error"} }
+        ]
+        for case in cases:
+            self.plugin_instance.handle_update_command = Mock(side_effect=HandlingException(case["status"]))
+            self.plugin_instance.on_api_command("update", {"pin": "r4"})
+            case["expectedMethod"].assert_called_with(case["expectedArgument"])
+
+    @patch("flask.jsonify")
+    def test_om_api_command__get_status_exception(self, jsonify_mock):
+        # Should respond with status false
```
RobinTail, line 1060:
```suggestion
        # Should respond with status false when handler raises
```

RobinTail (commented):
ok

PR #242, opened 2023-10-22
Feat: API v2

### LINKED ISSUE #164: API inconsistency and payload props naming

RobinTail:
Closes #164 

Introducing API parameter `version` or `v` (optional).
The default `version` falls back to `1` so the API keeps its original behaviour by default until the next major release.

## Changes of version 2

- When requesting an action on relay, now calling it `subject` instead of `pin`.
- When receiving the relay state, now calling it `status` instead of `result` or `active`.
- When requesting the state of disabled relay now responding with `400` HTTP code instead of `status: False`
- When updating or cancelling something, there is no more `status: "ok" | "error"`, the `status` now always means the relay state (boolean), while error response is supplied via HTTP code.

### Inputs

| Method | Inputs before | Inputs after |
|--------|--------|--------|
| update | `pin, target` | `subject, target` |
| getStatus | `pin` | `subject` |
| listAllStatus | — | — | 
| cancelTask | `subject, target, owner` | `subject, target, owner` |

### Outputs

| Method | Outputs before | Outputs after |
|--------|--------|--------|
| update | `result: bool, status: ok` | `status: bool` |
| getStatus | `status` | `status` |
| listAllStatus | `[active, id, name]` | `[status, id, name]` | 
| cancelTask | `status: ok` | `cancelled: bool` |

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -152,30 +152,32 @@ class OctoRelayPlugin(
     def on_api_command(self, command, data):
         # pylint: disable=too-many-return-statements
         self._logger.info(f"Received the API command {command} with parameters: {data}")
-        index = data.get("pin")
-        if command in [GET_STATUS_COMMAND, UPDATE_COMMAND] and index is None:
-            return flask.abort(400, description="Parameter pin is missing")
+        version = data.get("version") or data.get("v") or 1;
+        subject_param_name = "pin" if version == 1 else "subject"
+        subject = data.get(subject_param_name)
+        target = data.get("target")
+        if command in [GET_STATUS_COMMAND, UPDATE_COMMAND] and subject is None:
+            return flask.abort(400, description=f"Parameter {subject_param_name} is missing")
         if command == LIST_ALL_COMMAND: # API command to get relay statuses
             response = self.handle_list_all_command()
             self._logger.info(f"Responding to {LIST_ALL_COMMAND} command: {response}")
             return flask.jsonify(response)
         if command == GET_STATUS_COMMAND: # API command to get relay status
             try:
-                is_closed = self.handle_get_status_command(index)
+                is_closed = self.handle_get_status_command(subject)
             except HandlingException:
                 is_closed = False # todo should just abort in the next version
             self._logger.info(f"Responding to {GET_STATUS_COMMAND} command: {is_closed}")
             return flask.jsonify({"status": is_closed})
         if command == UPDATE_COMMAND: # API command to toggle the relay
-            target = data.get("target")
             try:
-                state = self.handle_update_command(index, target if isinstance(target, bool) else None)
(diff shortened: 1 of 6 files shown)


octoprint_octorelay/__init__.py
```diff
@@ -152,33 +152,55 @@ def handle_cancel_task_command(self, subject: str, target: bool, owner: str) ->
...
-        if command == LIST_ALL_COMMAND: # API command to get relay statuses
-            response = self.handle_list_all_command()
-            self._logger.info(f"Responding to {LIST_ALL_COMMAND} command: {response}")
+        version = data.get("version") or data.get("v") or 1
+        subject_param_name = "pin" if version == 1 else "subject"
+        subject = data.get(subject_param_name)
+        target = data.get("target")
+        if command in [GET_STATUS_COMMAND, UPDATE_COMMAND] and subject is None:
+            return flask.abort(400, description=f"Parameter {subject_param_name} is missing")
+        # API command to get relay statuses
```
RobinTail, line 161:
```suggestion
        # API command to list all the relays with their names and statuses
```

octoprint_octorelay/__init__.py
```diff
@@ -152,33 +152,55 @@ def handle_cancel_task_command(self, subject: str, target: bool, owner: str) ->
     def on_api_command(self, command, data):
         # pylint: disable=too-many-return-statements
         self._logger.info(f"Received the API command {command} with parameters: {data}")
-        index = data.get("pin")
-        if command in [GET_STATUS_COMMAND, UPDATE_COMMAND] and index is None:
-            return flask.abort(400, description="Parameter pin is missing")
-        if command == LIST_ALL_COMMAND: # API command to get relay statuses
-            response = self.handle_list_all_command()
-            self._logger.info(f"Responding to {LIST_ALL_COMMAND} command: {response}")
+        version = data.get("version") or data.get("v") or 1
```
RobinTail, line 155:
```suggestion
        version = int( data.get("version") or data.get("v") or 1 )
```

octoprint_octorelay/__init__.py
```diff
@@ -152,33 +152,55 @@ def handle_cancel_task_command(self, subject: str, target: bool, owner: str) ->
...
+        subject_param_name = "pin" if version == 1 else "subject"
```
RobinTail, line 156:
```suggestion
        subject_param_name = "pin" if version == 1 else "subject" # todo remove pin when dropping v1
```

RobinTail (commented):
LGTM

RobinTail (commented):
ok

RobinTail (commented):
ready

PR #257, opened 2024-02-12
🚀 4.1.0

### LINKED ISSUE #259: Specific Gcode commands for turning relays on or off
Currently the ony Gcode command available is for switching back and forth. This works in most cases, but in my opinion there are also many cases where it would be easier to just have a specific command for on and a specific command for off, instead of having to expect a specific state of the relais before executing the Gcode command. 

Example: I have been building an enclosure for my printer in the past few month which also uses buttons to execute commands and LEDs to give feedback. On of these buttons is for Play/Pause, which is supposed to be off when nothing is being printed, on while printing and blinking when a print has been paused for some reason. I had to use 2 relais and a blinker circuit to make this work, but if I stop the print while it's paused I have no way of making the second relais switch back to the default state, because there is no specific command for turning it off, only for switching it back and forth.

RobinTail:
Closes #259 

- #254 
- #256 
- #260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Version 4
 
+### 4.1.0
+
+- New feature: target for GCODE command (at-command):
+  - You can now switch the relay to the desired state;
+  - `@OCTORELAY r1 ON` — to switch the first relay on;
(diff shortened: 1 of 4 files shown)


RobinTail:
✅ QA passed

RobinTail (commented):
ready

Issue #255, opened 2024-02-04
Buttons not showing on Raspberry Pi 5


ionicaalexandru:
Hi,

did someone encounter this issue: octorelay buttons not showing in the navbar?
Im using octoprint 1.9.3 Python 3.11.2 on a Raspberry Pi 5 8Gb Board with default plugins installed

TNX

ckiefer:
Indeed. I have the same setup and I don't see the buttons either.

BroncoNS:
Exact same for me as well.

ionicaalexandru:
> @ionicaalexandru , @ckiefer , @BroncoNS ,
> 
> > README.md
> > After installing the plugin you need to configure it in order to see the control buttons in the navigation bar.
> 
> Did you configure the plugin and set some of the buttons "Active"?
> 
[...]

PR #271, opened 2024-05-19
Experiment: Using `gpiod` for driving pins

### LINKED ISSUE #270: BIGTREETECH BTT  Pi V1.2  The plugin is installed but it is not in the plugin manager.
On BIG TREE TECH BTT Pi V1.2 from debian 11 taken from github - CB1_Debian11_minimal_kernel5.16_20240319.img.xz     Octoprint 1.10 
The plugin is installed but it is not in the plugin manager. In protected mode, I see this plugin. What to do?

RobinTail:
Probably closes #270 and maybe will help with #255

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,7 +7,7 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-[mypy-RPi.*]
+[mypy-gpiod.*]
 ignore_missing_imports = True
 
 [mypy-flask.*]
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,9 +7,5 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-
-[mypy-gpiod.*]
-ignore_missing_imports = True
-
 [mypy-flask.*]
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,5 +7,9 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
+
+[mypy-gpiod.*]
+ignore_missing_imports = True
+
 [mypy-flask.*]
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,8 +7,6 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-[mypy-gpiod]
-disable_error_code = attr-defined
 
 [mypy-gpiod.*]
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,7 +7,6 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-
 [mypy-gpiod.*]
 ignore_missing_imports = True
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,6 +7,9 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
+[mypy-gpiod]
+disable_error_code = attr-defined
+
 [mypy-gpiod.*]
 ignore_missing_imports = True
 
--- a/octoprint_octorelay/driver.py
+++ b/octoprint_octorelay/driver.py
@@ -12,7 +12,7 @@ class Relay():
             "/dev/gpiochip0",
             consumer = "OctoRelay",
             config = {
-                pin: LineSettings(direction=Direction.OUTPUT)
+                pin: LineSettings(direction=Direction.OUTPUT, active_low=False)
             }
         )
         self.pin = pin # GPIO pin
--- a/octoprint_octorelay/driver.py
+++ b/octoprint_octorelay/driver.py
@@ -8,7 +8,13 @@ def xor(left: bool, right: bool) -> bool:
 
 class Relay():
     def __init__(self, pin: int, inverted: bool):
-        self.path = "/dev/gpiochip0"
+        self.request = request_lines(
+            "/dev/gpiochip0",
+            consumer = "OctoRelay",
+            config = {
+                pin: LineSettings(direction=Direction.OUTPUT)
+            }
+        )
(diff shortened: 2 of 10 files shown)


setup.py
```diff
@@ -51,7 +51,7 @@ def get_version_and_cmdclass(pkg_path):
 
 # Any additional requirements besides OctoPrint should be listed here
 # todo after dropping 3.7 remove typing-extensions (used by model.py and listing.py)
-plugin_requires = ["RPi.GPIO", "typing-extensions"]
+plugin_requires = ["gpiod", "typing-extensions"]
```
RobinTail, line 54: ⚠️ breaking: Requires: Python >=3.9.0

octoprint_octorelay/driver.py
```diff
@@ -1,15 +1,18 @@
 # -*- coding: utf-8 -*-
 from typing import Optional
-from RPi import GPIO
 
-# The driver operates BCM mode of pins enumeration
-GPIO.setmode(GPIO.BCM)
+# pylint: disable=no-name-in-module
```
RobinTail, line 4: this is due to unsupported python

octoprint_octorelay/driver.py
```diff
@@ -1,15 +1,18 @@
...
+from gpiod import request_lines, LineSettings # type: ignore[attr-defined]
+
+# pylint: disable=import-error
```
RobinTail, line 7: this is due to unsupported python

mypy.ini
```diff
@@ -7,7 +7,10 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-[mypy-RPi.*]
+[mypy-gpiod]
+disable_error_code = attr-defined
```
RobinTail, line 11:
```suggestion
```

octoprint_octorelay/driver.py
RobinTail:
```suggestion
```

octoprint_octorelay/driver.py
RobinTail:
```suggestion
```

octoprint_octorelay/driver.py
```diff
@@ -26,10 +25,14 @@ def open(self):
 
     def is_closed(self) -> bool:
         """Returns the logical state of the relay."""
-        GPIO.setwarnings(False)
-        GPIO.setup(self.pin, GPIO.OUT)
-        pin_state = bool(GPIO.input(self.pin))
-        GPIO.setwarnings(True)
+        request = request_lines(
```
RobinTail, line 28: DNRY: this can probably be reused by keeping the request as a single prop of Relay

RobinTail: addressed

mypy.ini
```diff
@@ -7,7 +7,8 @@ exclude = (?x)(
 [mypy-octoprint.*]
 ignore_missing_imports = True
 
-[mypy-RPi.*]
+
```
RobinTail, line 10:
```suggestion
```

setup.py
RobinTail: addressed in Readme and entrypoint

setup.py
```diff
@@ -35,7 +35,7 @@ def get_version_and_cmdclass(pkg_path):
 
 # The plugin's description. Can be overwritten within OctoPrint's internal data via __plugin_description__ in the plugin
 # module
-plugin_description = """A plugin that adds buttons to the navbar for toggling GPIO on the RPi. It can be used for turning relays on and off."""
+plugin_description = """A plugin that adds buttons to the navbar for toggling GPIO on Raspberry Pi. It can be used for turning relays on and off."""
```
RobinTail, line 38: not only Raspberry Pi now

octoprint_octorelay/driver.py
```diff
@@ -1,15 +1,20 @@
...
+from gpiod import request_lines, LineSettings
+from gpiod.line import Direction, Value
 
 def xor(left: bool, right: bool) -> bool:
     return left is not right
 
 class Relay():
     def __init__(self, pin: int, inverted: bool):
+        self.request = request_lines(
+            "/dev/gpiochip0",
```
RobinTail, line 12: this should be configurable

tests/test_driver.py
```diff
@@ -4,10 +4,17 @@
...
+line_mock = Mock()
+gpiod_mock.request_lines = Mock(return_value=line_mock)
+gpiod_mock.LineSettings = Mock(return_value="LineSettingsMock")
+sys.modules["gpiod"] = gpiod_mock
+line_mock.Direction = Mock()
+line_mock.Direction.OUTPUT = "OutputMock"
+line_mock.Value = Mock()
+line_mock.Value.ACTIVE = "ActiveMock"
+line_mock.Value.INACTIVE = "InactiveMock"
+sys.modules["gpiod.line"] = line_mock
```
RobinTail, line 17: this should be another variable

RobinTail: addressed in 8f839ee

RobinTail:
🚀 5.0.0a2

:x: not fixed

`request_lines` seems to be always pushing some initial value, it should not be used for reading the state

RobinTail:
It seems that this library won't work with existing code implementation treating each switch as a standalone operation.
Instead, the driver should either be a singleton, or the code should be refactored to create `Relay` instances once and then operate them as needed.
https://www.tomshardware.com/how-to/control-raspberry-pi-5-gpio-with-python-3

RobinTail:
I also found out, that there is a mode `Direction.AS_IS` that does not set the initial value.
Moreover, it's possible to set value on demand from that mode.
However, I'm not sure how'd it work if the pin is initially in Input mode.

RobinTail:
Alternatively, there is a much better documented library
https://gpiozero.readthedocs.io/en/latest/index.html
That is confirmed to work on Raspberry 5 (applicable to #255 )
But not sure if it's applicable for #270

RobinTail:
🚀 5.0.0a3

🆗 Fixed initial state
🆗 No conflicts this time
🆗 Switching works
❌ Polling thread does not seems to monitor changes correctly

RobinTail:
Won't work without much greater refactoring

PR #274, opened 2024-07-05
Migration to gpiozero

ch3p4ll3:
Migration to the [gpiozero ](https://gpiozero.readthedocs.io/en/latest/) library to make it compatible to the Raspberry Pi5 as well. This will resolve the problems from [issue#255](https://github.com/borisbu/OctoRelay/issues/255).


octoprint_octorelay/__init__.py
```diff
@@ -124,7 +125,7 @@ def handle_get_status_command(self, index: str) -> bool:
         settings = self._settings.get([index], merged=True) # expensive
         if not bool(settings["active"]):
             raise HandlingException(400)
-        return Relay(
+        return  Relay.get_or_create_relay(
```
RobinTail, line 128:
```suggestion
        return Relay.get_or_create_relay(
```

setup.py
```diff
@@ -51,7 +51,7 @@ def get_version_and_cmdclass(pkg_path):
 
 # Any additional requirements besides OctoPrint should be listed here
 # todo after dropping 3.7 remove typing-extensions (used by model.py and listing.py)
-plugin_requires = ["RPi.GPIO", "typing-extensions"]
+plugin_requires = ["gpiozero", "typing-extensions", "lgpio"]
```
RobinTail, line 54: Thank you @ch3p4ll3 !
Could you please clarify "lgpio": what's the purpose of it — I didn't find its usage in your changes.

ch3p4ll3: Hi, lgpio is used by gpiozero for building the [pin factory.](https://gpiozero.readthedocs.io/en/v1.6.0/api_pins.html) so there is no "direct" usage of it

RobinTail: So, `gpiozero` automatically detects that `lgpio` is installed and then use it as a pin factory, is that correct @ch3p4ll3 ?

RobinTail:
There is an issue in CI, @ch3p4ll3 with `lgpio`:

```
Building wheels for collected packages: lgpio
  Building wheel for lgpio (setup.py): started
  Building wheel for lgpio (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  Running setup.py clean for lgpio
  ╰─> [15 lines of output]
      running bdist_wheel
      running build
      running build_py
      running build_ext
      building '_lgpio' extension
      swigging lgpio.i to lgpio_wrap.c
      swig -python -o lgpio_wrap.c lgpio.i
      creating build
      creating build/temp.linux-x86_64-3.8
      gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Isrc -I/opt/hostedtoolcache/Python/3.8.18/x64/include/python3.8 -c lgpio_wrap.c -o build/temp.linux-x86_64-3.8/lgpio_wrap.o
      creating build/lib.linux-x86_64-3.8
      gcc -shared -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib build/temp.linux-x86_64-3.8/lgpio_wrap.o -L/opt/hostedtoolcache/Python/3.8.18/x64/lib -llgpio -o build/lib.linux-x86_64-3.8/_lgpio.cpython-38-x86_64-linux-gnu.so
      /usr/bin/ld: cannot find -llgpio: No such file or directory
      collect2: error: ld returned 1 exit status
      error: command 'gcc' failed with exit status 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for lgpio
Failed to build lgpio
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (lgpio)
```

Could you take a look on what it takes to fix it?

setup.py
ch3p4ll3: Yep, exactly

ch3p4ll3:
> There is an issue in CI, @ch3p4ll3 with installing `lgpio`:
> 
> ```
> Building wheels for collected packages: lgpio
>   Building wheel for lgpio (setup.py): started
>   Building wheel for lgpio (setup.py): finished with status 'error'
>   error: subprocess-exited-with-error
>   
>   × python setup.py bdist_wheel did not run successfully.
>   │ exit code: 1
>   Running setup.py clean for lgpio
>   ╰─> [15 lines of output]
>       running bdist_wheel
>       running build
>       running build_py
>       running build_ext
>       building '_lgpio' extension
>       swigging lgpio.i to lgpio_wrap.c
>       swig -python -o lgpio_wrap.c lgpio.i
>       creating build
>       creating build/temp.linux-x86_64-3.8
>       gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Isrc -I/opt/hostedtoolcache/Python/3.8.18/x64/include/python3.8 -c lgpio_wrap.c -o build/temp.linux-x86_64-3.8/lgpio_wrap.o
>       creating build/lib.linux-x86_64-3.8
>       gcc -shared -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib build/temp.linux-x86_64-3.8/lgpio_wrap.o -L/opt/hostedtoolcache/Python/3.8.18/x64/lib -llgpio -o build/lib.linux-x86_64-3.8/_lgpio.cpython-38-x86_64-linux-gnu.so
>       /usr/bin/ld: cannot find -llgpio: No such file or directory
>       collect2: error: ld returned 1 exit status
>       error: command 'gcc' failed with exit status 1
>       [end of output]
>   
>   note: This error originates from a subprocess, and is likely not a problem with pip.
>   ERROR: Failed building wheel for lgpio
> Failed to build lgpio
> ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (lgpio)
> ```
> 
> Could you take a look on what it takes to fix it?
> 
> That seems to be happening only in Python 3.7 and 3.8 — perhaps those versions are not supported?

Mmm interesting... Ok, I'll take a look

ch3p4ll3:
> There is an issue in CI, @ch3p4ll3 with installing `lgpio`:
> 
> ```
> Building wheels for collected packages: lgpio
>   Building wheel for lgpio (setup.py): started
>   Building wheel for lgpio (setup.py): finished with status 'error'
>   error: subprocess-exited-with-error
>   
>   × python setup.py bdist_wheel did not run successfully.
>   │ exit code: 1
>   Running setup.py clean for lgpio
>   ╰─> [15 lines of output]
>       running bdist_wheel
>       running build
>       running build_py
>       running build_ext
>       building '_lgpio' extension
>       swigging lgpio.i to lgpio_wrap.c
>       swig -python -o lgpio_wrap.c lgpio.i
>       creating build
>       creating build/temp.linux-x86_64-3.8
>       gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Isrc -I/opt/hostedtoolcache/Python/3.8.18/x64/include/python3.8 -c lgpio_wrap.c -o build/temp.linux-x86_64-3.8/lgpio_wrap.o
>       creating build/lib.linux-x86_64-3.8
>       gcc -shared -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.18/x64/lib build/temp.linux-x86_64-3.8/lgpio_wrap.o -L/opt/hostedtoolcache/Python/3.8.18/x64/lib -llgpio -o build/lib.linux-x86_64-3.8/_lgpio.cpython-38-x86_64-linux-gnu.so
>       /usr/bin/ld: cannot find -llgpio: No such file or directory
>       collect2: error: ld returned 1 exit status
>       error: command 'gcc' failed with exit status 1
>       [end of output]
>   
>   note: This error originates from a subprocess, and is likely not a problem with pip.
>   ERROR: Failed building wheel for lgpio
> Failed to build lgpio
> ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (lgpio)
> ```
> 
> Could you take a look on what it takes to fix it?
> 
> That seems to be happening only in Python 3.7 and 3.8 — perhaps those versions are not supported?

Ok, so I found out that unfortunately `lgpio` supports `python>=3.9`  [[1]](https://stackoverflow.com/questions/70903849/how-to-install-python3-lgpio-on-ubuntu-20-04-on-a-raspberry-pi) :(

RobinTail:
> Ok, so I found out that unfortunately lgpio supports python>=3.9 [[1]](https://stackoverflow.com/questions/70903849/how-to-install-python3-lgpio-on-ubuntu-20-04-on-a-raspberry-pi) :(

Ok. Should we utilize another pin factory or should we drop Python 3.7 and 3.8 in favour of `lgpio`? 
What do you think, @ch3p4ll3 ?

octoprint_octorelay/driver.py
```diff
@@ -40,8 +37,29 @@ def toggle(self, desired_state: Optional[bool] = None) -> bool:
...
+    def get_or_create_relay(cls, pin: int, inverted: bool):
+        to_return = None
+
+        for relay in cls.relays:
+            if relay.pin == pin:
+                if relay.inverted != inverted:
+                    relay.inverted = inverted
+
+                to_return = relay
+                break
```
RobinTail, line 58: perhaps we can just `return relay` here and thus get rid of `to_return`, what do you think?

octoprint_octorelay/driver.py
```diff
@@ -40,8 +37,29 @@ def toggle(self, desired_state: Optional[bool] = None) -> bool:
...
+
         return desired_state
+
+    @classmethod
+    def get_or_create_relay(cls, pin: int, inverted: bool):
+        to_return = None
+
+        for relay in cls.relays:
+            if relay.pin == pin:
+                if relay.inverted != inverted:
```
RobinTail, line 54:
```suggestion
                if relay.inverted is not inverted:
```

I heard it's better to use this syntax for comparing booleans

ch3p4ll3:
> > Ok, so I found out that unfortunately lgpio supports python>=3.9 [[1]](https://stackoverflow.com/questions/70903849/how-to-install-python3-lgpio-on-ubuntu-20-04-on-a-raspberry-pi) :(
> 
> Ok. Should we utilize another pin factory or should we drop Python 3.7 and 3.8 in favour of `lgpio`? What do you think, @ch3p4ll3 ?
> 
> In case you believe `lgpio` is a must, you can adjust this block in the CI workflow:
> 
> https://github.com/borisbu/OctoRelay/blob/7c15a7077766a7256b77603cfc00b91284b57ca6/.github/workflows/CI.yaml#L54-L83
> 
> and this constant in the main file:
> 
> https://github.com/borisbu/OctoRelay/blob/7c15a7077766a7256b77603cfc00b91284b57ca6/octoprint_octorelay/__init__.py#L374
> 
> An example here: #215 — I'm actually looking forward to drop at least 3.7.

From the tests I have done on the Pi5, `lgpio` is the only pin factory that works, other pin factories are not compatible with the new GPIO hardware. This would make `lgpio` the **only** option available for raspberry Pi 5 if we want to use this approach.

As far as dropping python 3.7/8, I don't know, the problem is that right now octoprint still supports both python 3.7 and 3.8. If we have to drop support, for some it might be necessary to upgrade python to at least 3.9. Imho this decision should not be up to me but from the repo owners

RobinTail:
> lgpio is the only pin factory that works
> lgpio the only option available for raspberry Pi 5

Thank you for making that research! Highly appreciate it.

> octoprint still supports both python 3.7 and 3.8
> this decision should not be up to me but from the repo owners

I'm OK with it. [Python 3.7 is EOL, 3.8 is about to](https://devguide.python.org/versions/)
Plugins may have more narrow system requirements than the primary application in order to run their features — this is fine.
We can release a new major version with a breaking change to Python compatibility in order to offer the Raspberry 5 support.

RobinTail:
The installation was successful, @ch3p4ll3 .
Now unit tests need to be updated.
You can run them locally using the following command from the `tests` directory:

```sh
python -m unittest test*.py
```

ch3p4ll3:
> The installation was successful, @ch3p4ll3 . Now unit tests need to be updated. You can run them locally using the following command from the `tests` directory:
> 
> ```shell
> python -m unittest test*.py
> ```
> 
> Or just follow the CI steps. You'd need to mock `gpiozero` instead of `RPi.GPIO` in tests. Or find the CI execution log here: https://github.com/borisbu/OctoRelay/actions/runs/9820228078/job/27115376938?pr=274
> 
> Let me know if you need help.

Hi, I'm going to try to update them tomorrow, maybe I'll need some help since I've never used Mock and from what I've seen they don't look very simple to me lmao

RobinTail:
Ok, @ch3p4ll3 .
I'm going to collaborate with you on this PR: I will do several commits.
Please make sure to `git pull --rebase` before doing `git push` of your future commits.

ch3p4ll3:
Hi @RobinTail , I tried to update test_driver, could u please take a look, idk if that's good enought tbh. If that's ok I'll start working also on test_init

tests/test_driver.py
```diff
@@ -19,40 +15,33 @@
 
 class TestRelayDriver(unittest.TestCase):
     def test_constructor(self):
-        GPIO_mock.setmode.assert_called_with("MockedBCM")
-        relay = Relay(18, True)
+        relay = Relay.get_or_create_relay(1, True, MockFactory())
```
RobinTail, line 18: 1. this test is for the class constructor, the static method you made should be tested separately
2. please preserve the pin numbers of the tests, only change logic, not data (here and below)

```suggestion
        relay = Relay(18, True, MockFactory())
```

tests/test_driver.py
```diff
@@ -19,40 +15,33 @@
...
         self.assertIsInstance(relay, Relay)
-        self.assertEqual(relay.pin, 18)
+        self.assertEqual(relay.pin, 1)
```
RobinTail, line 20:
```suggestion
        self.assertEqual(relay.pin, 18)
```

please preserve the pin numbers of the tests, only change logic, not data (here and below)

tests/test_driver.py
```diff
@@ -19,40 +15,33 @@
...
         self.assertTrue(relay.inverted)
 
     def test_serialization(self):
-        relay = Relay(18, True)
+        relay = Relay.get_or_create_relay(2, True, MockFactory())
```
RobinTail, line 24:
```suggestion
        relay = Relay.get_or_create_relay(18, True, MockFactory())
```

tests/test_driver.py
```diff
@@ -19,40 +15,33 @@
...
         serialization = f"{relay}"
-        self.assertEqual(serialization, "Relay(pin=18,inverted=True,closed=True)")
+        self.assertEqual(serialization, "Relay(pin=2,inverted=True,closed=True)")
```
RobinTail, line 26:
```suggestion
        self.assertEqual(serialization, "Relay(pin=18,inverted=True,closed=True)")
```

RobinTail:
@ch3p4ll3 , well, it's kinda OK, but:
- could you please avoid changing the test data (pins) in the tests

I left a couple notes.
But in general, that could work. In the test of main, the whole driver is mocked, so it should be plausible. 
Thank you

tests/test_driver.py
```diff
@@ -61,13 +50,10 @@ def test_is_closed(self):
             { "mocked_state": 1, "inverted": True, "expected_relay_state": False },
             { "mocked_state": 0, "inverted": True, "expected_relay_state": True },
         ]
-        for case in cases:
-            GPIO_mock.input = Mock(return_value=case["mocked_state"])
-            relay = Relay(18, case["inverted"])
+        for case_id, case in enumerate(cases):
+            relay = Relay.get_or_create_relay(case_id + 10, case["inverted"], MockFactory())
```
RobinTail, line 54: consider keeping the `Relay()` (constructor) here to reduce this magic with `+10`.
That allows to keep each test case independent from cache, thus it does not need different pins and you could keep the original `18` for them all

RobinTail: the more straightforward the test data is — it's better

RobinTail: Yes, so that you can minimize the changes of your PR just by keeping the `Relay()` and testing that new static method separately.

tests/test_driver.py
```diff
@@ -61,13 +50,10 @@ def test_is_closed(self):
...
+            relay.relay.value = case["mocked_state"]
```
RobinTail, line 55: this should probably be an argument of `MockFactory()`

octoprint_octorelay/driver.py
```diff
@@ -1,21 +1,27 @@
...
 
-# The driver operates BCM mode of pins enumeration
-GPIO.setmode(GPIO.BCM)
-
-def xor(left: bool, right: bool) -> bool:
-    return left is not right
 
 class Relay():
-    def __init__(self, pin: int, inverted: bool):
+    relays = []
```
RobinTail, line 8: consider `cache`

tests/test_driver_get_or_create.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_get(self):
+        relay1 = Relay.get_or_create_relay(20, False, MockFactory())
+        relay2 = Relay.get_or_create_relay(20, False, MockFactory())
+        
+        self.assertIs(relay1, relay2)
+        self.assertEqual(len(Relay.relays), 1)
+        
+        relay3 = Relay.get_or_create_relay(21, True, MockFactory())
+        self.assertNotEqual(relay1, relay3)
+        self.assertEqual(len(Relay.relays), 2)
```
RobinTail, line 26: I think this test is enough.
Can it live inside the existing `test_driver.py`?

tests/test_driver_get_or_create.py
```diff
@@ -0,0 +1,50 @@
...
+# pylint: disable=wrong-import-position
+from octoprint_octorelay.driver import Relay
+
+# avoid keeping other modules automatically imported by this test
+del sys.modules["octoprint_octorelay"]
+del sys.modules["octoprint_octorelay.migrations"]
+del sys.modules["octoprint_octorelay.task"]
+
+class TestRelayDriver(unittest.TestCase):
+    def test_get(self):
```
RobinTail, line 17: can these tests be moved into the existing `test_driver.py`?

RobinTail: I also think it can be reduced into one test by just making the second relay inverted

```py
        relay2 = Relay.get_or_create_relay(20, True, MockFactory()) # inverted
```

tests/test_driver.py
RobinTail: fixed

tests/test_driver.py
RobinTail: fixed

tests/test_driver.py
RobinTail: fixed

tests/test_driver.py
RobinTail: fixed

RobinTail:
Something fails in running the tests — I think it can be resolved by moving `test__get` into the existing `test_drivier.py` suit.
You can make it a single test for the static method as advised above, @ch3p4ll3 
You're doing great 👍🏽

tests/test_driver.py
RobinTail: fixed

octoprint_octorelay/driver.py
ch3p4ll3: I don't think you can move the list inside the init, From what I know inside an object you can't create a new object, that's why I had done the class method. 

I created the `get_or_create` method because if you try to create multiple LED instances with the same pin, gpiozero raises an exception. There is a possibility to avoid this by using gpiozero's `.close()` function, but this resets the pin state to its initial state.

ch3p4ll3:
Fixed also `test_init.py` by adding a Mock for `get_or_create_relay`

octoprint_octorelay/driver.py
RobinTail: I meant the name of the property. So it could be `Relay.cache` instead of `Relay.relays`

RobinTail: But it's fine as it for now

RobinTail:
@ch3p4ll3 the unit tests passed — awesome!

A couple refinement needed to fix the found code style issues:

```
octoprint_octorelay/driver.py:3: error: Skipping analyzing "gpiozero.pins.lgpio": module is installed, but missing library stubs or py.typed marker  [import-untyped]
octoprint_octorelay/driver.py:3: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
octoprint_octorelay/driver.py:4: error: Skipping analyzing "gpiozero": module is installed, but missing library stubs or py.typed marker  [import-untyped]
octoprint_octorelay/driver.py:8: error: Need type annotation for "relays" (hint: "relays: list[<type>] = ...")  [var-annotated]
Found 3 errors in 1 file (checked 8 source files)
```

### Regarding the first one

In the file "mypy.ini" find the following block:
```ini
[mypy-RPi.*]
ignore_missing_imports = True
```

and replace "RPi" with "gpiozero"

### For the second

Add `: list[Relay]` next to the `relays` in `Relay` class.

That should fix it.

RobinTail:
Awesome. Last minor issues, @ch3p4ll3 : 

```
************* Module octoprint_octorelay.driver
octoprint_octorelay/driver.py:45:0: C0303: Trailing whitespace (trailing-whitespace)
************* Module test_driver
tests/test_driver.py:70:0: C0303: Trailing whitespace (trailing-whitespace)
tests/test_driver.py:77:0: C0303: Trailing whitespace (trailing-whitespace)
tests/test_driver.py:4:0: W0611: Unused Mock imported from unittest.mock (unused-import)
```

tests/test_driver.py
```diff
@@ -3,11 +3,7 @@
 import sys
 from unittest.mock import Mock
```
RobinTail, line 4: this import is no longer used, should be removed
```suggestion
```

RobinTail:
Brilliant, @ch3p4ll3 , thank you!
I'm going to merge it to develop and publish a dev-preview so that you could test in on Raspberry Pi 5 and me on Raspberry Pi 4.

RobinTail (approved):

RobinTail:
@ch3p4ll3 , see: https://github.com/borisbu/OctoRelay/pull/280#issuecomment-2218107437

PR #281, opened 2024-07-10
fix driver for automatic pin factory

ch3p4ll3:


ch3p4ll3:
Hi, this should fix the issues of yesterday. I tried it on my Pi5 with Raspbian and it works. 

Btw, it seems that gpiozero creates a cache(?) file, at the beginning I got this error: `py.warnings - WARNING - /home/pi/OctoPrint/venv/lib/python3.11/site-packages/gpiozero-2.0.1-py3.11.egg/gpiozero/devices.py:300: PinFactoryFallback: Falling back from lgpio: [Errno 2] No such file or directory: '.lgd-nfy-3'`.

I discovered that this is a permissions problem in the folder where octoprint is started from (being started with `systemd` I think some folder in the root). To fix this I had to add `WorkingDirectory=<your octoprint folder>`, eg: `WorkingDirectory=/home/pi/OctoPrint` to the file `/etc/systemd/system/octoprint.service` in the section `Service`. Then I reloaded the systemd services: `sudo systemctl daemon-reload` and finally restarted octoprint: `sudo systemctl restart octoprint`

This evening I'll try it on my Pi3 B+ with OctoPi. Let me know if you encounter some errors

RobinTail (approved):
OK 👍🏽

PR #289, opened 2024-07-12
🚀 5.0.1

RobinTail:
- #288 
- #290 
- #291 
- #292

### DIFF
--- a/octoprint_octorelay/__init__.py
+++ b/octoprint_octorelay/__init__.py
@@ -371,7 +371,6 @@ class OctoRelayPlugin(
                 self.update_ui()
                 break
 
-__plugin_pythoncompat__ = ">=3.9,<4"
 __plugin_implementation__ = OctoRelayPlugin()
 
 __plugin_hooks__ = {
@@ -384,4 +383,4 @@ __plugin_hooks__ = {
 }
 
 # pylint: disable=wrong-import-position
-from ._version import __version__
+from ._version import __version__, __plugin_pythoncompat__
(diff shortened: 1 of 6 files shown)


RobinTail:
Good thing:
- It really prevents installing with Python 3.8

```
ERROR: Package 'octorelay' requires a different Python: 3.8.19 not in '<4,>=3.9'
```

Bad thing:
- Octorelay somehow does not recognize the compatibility string and disables the plugin

```
WARNING - Plugin OctoRelay (5.0.1rc1) is not compatible to Python 3.9.2 (compatibility string: >=2.7,<3).
```

octoprint_octorelay/__init__.py
```diff
@@ -384,4 +383,4 @@ def input_polling(self):
 }
 
 # pylint: disable=wrong-import-position
-from ._version import __version__
+from ._version import __version__, __plugin_pythoncompat__
```
RobinTail, line 386: this does not work for OctoRelay

octoprint_octorelay/__init__.py
```diff
@@ -371,7 +371,8 @@ def input_polling(self):
                 self.update_ui()
                 break
 
-__plugin_pythoncompat__ = ">=3.9,<4"
+# pylint: disable=wrong-import-position
+from ._version import __version__, __plugin_pythoncompat__
```
RobinTail, line 375: does not work this way for OctoPrint

RobinTail:
qa passed

PR #293, opened 2024-07-13
Avoid excessive switches on startup

RobinTail:
If relay is inverted, `gpiozero` sets its initial value to `false`, which turns it on, because `false` is the default initial value in docs:

https://gpiozero.readthedocs.io/en/latest/api_output.html#led

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 5
 
+### 5.0.2
+
+- Fixed excessive switches for inverted relays on startup.
+
 ### 5.0.1
 
 - Preventing installations on environments having unsupported Python versions:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 5.0.2
 
-- Fixed excessive switches for inverted relays on startup.
+- Initial state for relays is `OFF` — fixes excessive switches on startup in most cases.
 
 ### 5.0.1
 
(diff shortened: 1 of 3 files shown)


RobinTail (commented):
should be fine

octoprint_octorelay/driver.py
```diff
@@ -11,7 +11,7 @@ class Driver():
     def __init__(self, pin: int, inverted: bool, pin_factory=None):
         self.pin = pin # GPIO pin
         self.inverted = inverted # marks the relay as normally closed
-        self.handle = LED(pin, pin_factory=pin_factory)
+        self.handle = LED(pin, pin_factory=pin_factory, initial_value=None)
```
RobinTail, line 14: or maybe it should be `not inverted`...

RobinTail:
I found that it was the same in v4, we didn't set initial value in RPI.GPIO.setup

octoprint_octorelay/driver.py
RobinTail: tested locally:
- `None` works good for inverted relays
- `inverted` works good for all relays

```suggestion
        self.handle = LED(pin, pin_factory=pin_factory, initial_value=inverted)
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 5
 
+### 5.0.2
+
+- Fixed excessive switches for inverted relays on startup.
```
RobinTail, line 7:
```suggestion
- Initial state for relays is `OFF` — fixes excessive switches on startup in most cases.
```

RobinTail (commented):
ok

Issue #299, opened 2024-08-07
Physical Button Plugin broken


ccle610:
Plugin Version
5.0.1

Question
So far I have used "Physical Button"a GPIO Input via Hardware Button to toggle the light ( =>Action: GPIO toggle).

The output is also used by the Octorelay plug-in to control the light. This worked well for a long time. Unfortunately, the button no longer reacts as long as the Octo Relay plug-in is running. Without the Octorelay plug-in, everything works perfectly!
I also tried the Workaround with GCODE in Physical Button, but also this does not work.

Is this a topic for here, or for Physicl Button?

Additional context
Physical Button Version 0.6.2

RobinTail:
yeah, I made a notice about it...

RobinTail:
Here is the solution for that problem, @ccle610 : 

https://github.com/borisbu/OctoRelay/issues/263#issuecomment-2241546096

ccle610:
Thx so much for this hint !

ccle610:
Thx so much for this hint !

RobinTail:
There is probably a way to bypass the lock (exclusive pin reservation) established by `gpiozero` with a single line, that I'm currently investigating here, trying to make a cleanest solution possible:
https://github.com/borisbu/OctoRelay/pull/315

RobinTail:
Compatibility will be fixed in v5.1.0

Issue #263, opened 2024-04-03
A way to use physical buttons


OdinKadaver05:
I'm currently trying to make build an enclosure for my MK3s which also utilizes physical buttons for power etc. . Since the option of using physical buttons isn't integrated into OctoRelay I use the Physical Buttons add-on to execute Gcode-commands which then are picked up by Octorelay and then executed by the relais. This seemed to work fine until I noticed that apparently Octoprint doesn't just turn off the Gcode terminal when no printer is connected, it also disables the ability to execute Gcode commands from every other source, which is kindof problematic if I want to use this for a power button that turns off the printer but keeps running.

Long story short: it would be great if either inputs could be directly processed by octorelay or if there would be a simple way to use what physical buttons calls system commands. Some other solution to this is also fine with me

RobinTail:
If I understand correctly, you'd like to be able to configure additional GPIO pin for relay so that the plugin could read its state and interpret it somehow as a signal for switching the relay. Does that appear to be correct, @OdinKadaver05 ?

RobinTail:
Similar issue (resolved): https://github.com/borisbu/OctoRelay/issues/36
Similar issue (closed): https://github.com/borisbu/OctoRelay/issues/60

RobinTail:
Check this out, @OdinKadaver05 : 
https://github.com/borisbu/OctoRelay/issues/36#issuecomment-928911843

Could that approach work for you?

RobinTail:
Physical Button plugin mention in that thread:
https://github.com/LuxuSam/PhysicalButton

OdinKadaver05:
@RobinTail  Yes and no. In fact I'm already using that Plugin for controlling my fan, the lights and a play/pause button, but my problem is that I also want to control the power of my printer this way, while being able to also control it over the Webinterface and OctoApp.

I've hooked the relay for the PSU up and the automatic stuff (such as "turn relay on when server starts") works fine, but since the only way that I can get the Physical Button Plugin to send commands to OctoRelay is with the Gcode option in Physical Button, this doesn't work unless my printer is already powered on due to octoprint (from what I can tell) not processing Gcode commands while no printer is connected, even if the commands are from Plugin to Plugin.

![20240404_234549](https://github.com/borisbu/OctoRelay/assets/61711559/9585025e-e732-41ec-b3f1-247c0b72da1d)
[...]

PR #315, opened 2024-10-12
Feat: no pin lock, avoid exclusive reservations

RobinTail:
Regarding https://github.com/borisbu/OctoRelay/issues/263#issuecomment-2241546096
Fixes https://github.com/borisbu/OctoRelay/issues/299

Opened a discussion here: https://github.com/gpiozero/gpiozero/discussions/1175

Could someone help me to test this branch on their Raspberry Pi running Bookworm, please?
Let me know and I'll publish a prerelease.

### DIFF
--- a/octoprint_octorelay/driver.py
+++ b/octoprint_octorelay/driver.py
@@ -12,6 +12,8 @@ class Driver():
         self.pin = pin # GPIO pin
         self.inverted = inverted # marks the relay as normally closed
         self.handle = LED(pin, pin_factory=pin_factory, initial_value=inverted)
+        # release immediately, avoid lock, allow physical buttons to operate same relays:
+        self.handle.pin_factory.release_pins(self.handle, self.pin)
 
     def __repr__(self) -> str:
         return f"{type(self).__name__}(pin={self.pin},inverted={self.inverted},closed={self.is_closed()})"


octoprint_octorelay/driver.py
```diff
@@ -12,6 +12,8 @@ def __init__(self, pin: int, inverted: bool, pin_factory=None):
         self.pin = pin # GPIO pin
         self.inverted = inverted # marks the relay as normally closed
         self.handle = LED(pin, pin_factory=pin_factory, initial_value=inverted)
+        # release immediately, avoid lock, allow physical buttons to operate same relays:
+        self.handle.pin_factory.release_pins(self.handle, self.pin)
```
RobinTail, line 16: @ch3p4ll3 , do you know if it's possible somehow to avoid reservation at all?
I could not find any useful information about that problem, except the comment in `Factory` code:
https://gpiozero.readthedocs.io/en/latest/_modules/gpiozero/pins.html#Factory

```
Descendents *may* override the following methods, if applicable:
* :meth:`reserve_pins`
```

So that theoretically, a custom Pin Factory could have no-op `reserve_pins()` method.
But the problem is that we're using `None` for `pin_factory` relying on internal election between `lgpio` and `RPI.GPIO`.

I tested this line on my Pi 4, and it works, restoring previous operation for the same relay by a Physical Button plugin (it uses RPI.GPIO), though, the line itself does not look much pretty to me.

RobinTail: Election's happening in `_default_pin_factory()` called by public `ensure_pin_factory` of the `Device` class
https://gpiozero.readthedocs.io/en/latest/_modules/gpiozero/devices.html#Device

RobinTail: Or maybe we could replace `LED` with a composite device, that
1. inits pin factory,
2. extends it somehow or overrides its `reserve_pins` method with no-op,
3. creates actual `LED` instance
That's what I'm thinking about.
Though, perhaps it's not much cleaner than the line that releases afterwards

RobinTail: Ah... it's actually static

```py
@staticmethod
    def ensure_pin_factory():
```

so that we could probably call `Device.ensure_pin_factory()` without instantiating it!

RobinTail: If I make a build, could you check that it works on Raspberry Pi 5 (general startup, relay switching), @ch3p4ll3 ?

ch3p4ll3: > If I make a build, could you check that it works on Raspberry Pi 5 (general startup, relay switching), @ch3p4ll3 ?

Yep, no problem

RobinTail:
Gonna try it in 5.1.0rc2

PR #331, opened 2024-12-04
Upgrading all dependencies.

RobinTail:


ui/yarn.lock
RobinTail, line 1: can enforce vite to v6 in order to deduplicate esbuild

RobinTail (commented):
ok

PR #348, opened 2025-05-04
UI: Upgrading all dependencies

RobinTail:

PR #345, opened 2025-04-23
Supporting Python 3.13

RobinTail:
OctoPrint 1.11 now supports Python 3.13
https://github.com/OctoPrint/OctoPrint/releases/tag/1.11.0

PR #349, opened 2025-05-04
CI: downgrading `setuptools`

RobinTail:
There was a breaking change to setuptools, that affects the CI in #345 and #348 
Proper fix requires investigation and migrating to a different setup approach


RobinTail (commented):
should also fix release workflow I think

will figure that out later

Issue #354, opened 2025-06-19
Initial value should reflect the startup value (when no delay)


cesaregiannetti:
The `initial_value` configured on gpiozero `LED` class is OFF.
STARTUP value is applied afterwards.
This is an issue when restarting OctoPrint, because the pins are momentarily driven OFF.

I have a power supply controller connected to a GPIO pin that should be ON while OctoPrint is running, but I can't restart the server (for example after upgrading) because the power is cut after restart (it doesn't turn on again, it can only turn off).

I would like the `LED` to have an `initial_value` equal to startup value when the `delay` is 0.

On this matter, I have another suggestion: `LED` accepts `active_high` parameter, which can be set to `not inverted`, so that XORing with inverted can be avoided when reading or writing. I'll try to issue a PR

cesaregiannetti:
I found a very simple and neat solution: `initial_value=None`.
I tested it successfully for my use case and I believe this is the intended use of the parameter.
gpiozero documentation states: 
> If [None](https://docs.python.org/3.9/library/constants.html#None), the LED will be left in whatever state the pin is found in when configured for output

Moreover, this doesn't need the dirty trick of fetching the STARTUP value (thankfully, because I didn't understand which call to `ensure` actually instantiate the LED class).

Now #355 becomes relevant, because it conflicts with the change I want to make

PR #355, opened 2025-06-19
Use gpiozero active_high for inverted output

cesaregiannetti:
Built-in active_high property of LED (and Output in general) handles the inverted output already, simplifying the access.


RobinTail:
@cesaregiannetti , is it related to the #354 ?

cesaregiannetti:
> @cesaregiannetti , is it related to the #354 ?

Not really, I mentioned this PR in the issue because I saw the possibility while investigating the issue, but PR 355 does not fix issue 354. I don't know how to fix issue 354  in the right way, I don't know python and web, but I might find some time in the future to do so

cesaregiannetti:
@RobinTail I found a neat solution for #354, which is setting `initial_value=None`.
I want to open a pull request, do you like me to base it on develop or on cesaregiannetti:active_high ?

tests/test_driver.py
```diff
@@ -75,7 +75,7 @@ def test_ensure(self):
         self.assertFalse(relay1.inverted)
 
         # Test retrieving the existing relay with the same pin and inversion
-        relay2 = Driver.ensure(17, True, MockFactory())
+        relay2 = Driver.ensure(17, False, MockFactory())
```
RobinTail, line 78: @cesaregiannetti , the purpose of this line is to ensure that even with inversion (`True`) the `ensure()` method does not create a new instance, but reuses the existing one.
It's required for the cases when user changes the corresponding setting.
Please revert the change of this line

```suggestion
        relay2 = Driver.ensure(17, True, MockFactory())
```

cesaregiannetti: I think you're talking about line 83, where relay3 is inverted, whereas here the comment says
> Test retrieving the existing relay with the same pin and inversion

so I think relay2 is actually testing just the cache

cesaregiannetti: I also adjusted the other tests to reflect the change. there's no need to have pin state and relay state anymore, as they are the same, thanks to gpiozero interface, but I kept them for clarity

RobinTail: ok

Issue #373, opened 2026-04-27
Support OctoPrint 2.0


RobinTail:
First RC release: https://github.com/OctoPrint/OctoPrint/releases/tag/2.0.0rc1

Migration guide: https://docs.octoprint.org/en/dev/plugins/migration_2_0_0.html

Compatiblity validation tool: https://github.com/jacopotediosi/octoscanner

New TOML-based isolated setup approach: https://docs.octoprint.org/en/dev/plugins/migration_pyproject_toml.html#sec-plugins-pyproject-toml

RobinTail:
✅ ran the compatibility validation tool. No issues against 2.0.0, but there are several concerns towards the future 2.1.0 compatibility ⚠️ 
UPD: both addressed in 5.4.0

RobinTail:
✅ the migration guide scanned against the codebase: no changes needed
https://docs.octoprint.org/en/dev/plugins/migration_2_0_0.html

RobinTail:
2.0rc3 compatible so far, in #381

PR #380, opened 2026-05-02
fix(template): Enabling exclicit autoescape.

RobinTail:
Related to #373 

https://docs.octoprint.org/en/main/plugins/mixins.html#octoprint.plugin.TemplatePlugin.is_template_autoescaped

It's already enabled implicitly since OctoPrint 1.13 but this PR makes it explicit.
See https://community.octoprint.org/t/how-do-i-improve-my-plugins-security-by-enabling-autoescape/61067


RobinTail (commented):
ok

PR #393, opened 2026-08-22
fix(template): label keyword highlighting

RobinTail:
After #380 this thing doesn't look right

Screenshot 2026-08-22 at 22 41 44


RobinTail:
fixed

Screenshot 2026-08-22 at 22 48 12

PR #398, opened 2026-09-15
Do not set initial value

cesaregiannetti:
initial_value property of gpiozero LED class was set to "off". This is an issue, during OctoPrint restart, for relays controlling the printer power or even the OctoPrint itself (such as in my case) because the plugin turns the pin off momentarily, cutting the power. Using "None", the LED will be left in whatever state the pin is found in when configured for output (quoted from gpiozero LED documentation) and the initial value desired by the user will be applied immediately after inizialization by the "on Startup" event, for a seamless UX. #354


octoprint_octorelay/driver.py
```diff
@@ -11,7 +11,7 @@ class Driver():
     def __init__(self, pin: int, inverted: bool, pin_factory=None):
         self.pin = pin # GPIO pin
         self.inverted = inverted # marks the relay as normally closed
-        self.handle = LED(pin, pin_factory=pin_factory, initial_value=inverted)
+        self.handle = LED(pin, pin_factory=pin_factory, initial_value=None)
```
RobinTail, line 14: thank you @cesaregiannetti .
did you try it?
Cause I remember I had some issue with this value [earlier](https://github.com/borisbu/OctoRelay/pull/293/changes/17b857ed49f41b135b774c7ead62ed9f4d6ff64d#r1676863393).
But maybe it's alright now. I'm gonna try it

cesaregiannetti: Yes, I've been using this for months now

RobinTail:
@cesaregiannetti I tested your approach. Here is what I found out. 

Regardless of the `initial_value` relays turn OFF when restarting OctoPrint. This seems to be caused by GPIOZERO.

This is a deeper analysis by AI: 

> **Shutdown: gpiozero turns the pins into inputs.** gpiozero registers an `atexit` hook (`devices.py:_devices_shutdown`). It calls `Device.pin_factory.close()`, which runs `pin.close()` on every pin the factory has seen. `LGPIOPin.close()` calls `lgpio.gpio_claim_input(..., SET_PULL_NONE)`, so each relay pin becomes an input with no pull and stops driving the relay, which usually drops. Calling `release_pins()` in `driver.py:16` doesn't prevent this. It only removes the *reservation*: `dev.close()` is skipped, but `factory.close()` still resets the pin.

Tested on pins 12 (inverted) and 4 (not inverted, transistor switch), having "on Startup=Skip" in the settings.
They both go OFF when I click "Restart OctoPrint" and then go back ON when it restarts.

So, I'd say that unfortunately the suggested change does not fix the stated problem (at least for particular cases). Correct me if I'm doing it wrong.

